PR #25, opened 2014-11-18
Add ipv6 support (DO NOT MERGE YET)

nojb:
See mirage/mirage-tcpip#70 and mirage/mirage#319.  I did just enough to be able to run the cohttp examples over ipv6.

Right now you are forced to listen for connections over both IPv4 and IPv6 together (see function `serve`).  We can fix this by making the type of `mode` have variants `TCPv4 of int` and `TCPv6 of int`.

### DIFF
--- a/lib/conduit_mirage.ml
+++ b/lib/conduit_mirage.ml
@@ -140,7 +140,7 @@ module type VCHAN_PEER = PEER
   with type uuid = string
    and type port = vchan_port
 
-module Make(S:V2_LWT.STACK)(V:VCHAN_PEER) = struct
+module Make(S:V1_LWT.NETSTACK)(V:VCHAN_PEER) = struct
 
   module Flow = Make_flow(S.TCPV4)(S.TCPV6)(V.Endpoint)
   type +'a io = 'a Lwt.t
--- a/lib/conduit_mirage.ml
+++ b/lib/conduit_mirage.ml
@@ -38,7 +38,7 @@ module type VCHAN_FLOW = V1_LWT.FLOW
   with type error := unknown
 
 (** All the possible connection types supported *)
-module Make_flow(S:V1_LWT.TCPV4)(V:VCHAN_FLOW) = struct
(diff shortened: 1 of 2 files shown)


avsm:
Would you have a rebased version of this for the latest tree?

nojb:
I can rebase but this would require to have STACKV6 in `mirage-tcpip`, which used to exist but not anymore.  If you want to, I can revive STACKV6 and then rebase this to get it back into working shape.

samoht:
This PR would be solved if we manage to get a `V1_LWT.STACK` module in `mirage-types`. cc/ @nojb do you still have patches laying around for that?

nojb:
Do you mean `V1_LWT.STACK` to be a common signature for a stack that handles **either** `V4` **or** `V6` ? or a stack that handles **both** `V4` and `V6` ?  I think both would be easy to do.  The reason I didn't move forward with this patch is that the current `STACKV4` seems to be very much in need of a redesign, and duplicating its functionality for the `V6` version (under either of the above escenarios) didn't seem a very good idea.

Having said that, I have the code somewhere because I needed it in order to get `mirage-www` working over `IPv6`, so I could revive it if you want ...

samoht:
> Do you mean V1_LWT.STACK to be a common signature for a stack that handles either V4 or V6 

yea. I agree that the `STACKV4` needs a complete redesign, I think both options are ok and are better than merging this pull request (as it adds again one new argument to the functor).

nojb:
This was written as a quick hack to test the IPv6 implementation in `mirage-tcpip` some time ago, it is most surely in need of updating, maybe of a complete rewrite.

As for the status of IPv6, there is an implementation of it in `mirage-tcpip` (probably needing more testing) but AFAIK it is not integrated into the rest of the Mirage infrastructure (the `STACK` implementation, `conduit`, etc.)

nojb:
Not relevant anymore.

dinosaure:
Yes, due to the ability to define by yourself the type `'edn`, it's possible to return an `IPv6`. Then, a _protocol_ which handles IPv6 must exists (such the tcp/ip UNIX stack).

PR #37, opened 2015-01-13
Mirage TLS support

talex5:
The first patch makes use of the new `error_message` function in https://github.com/mirage/mirage/pull/346 to simplify error handling for flows.

The second patch adds TLS support. It's used like this:

        let mode = `TLS (tls_config, `TCP (`Port 8443)) in
        Conduit1.serve ~ctx ~mode (H.listen spec)

i.e. you pass in a TLS configuration and an underlying conduit specification and it wraps it.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,4 +1,4 @@
-OPAM_DEPENDS="sexplib ipaddr cstruct stringext uri vchan"
+OPAM_DEPENDS="sexplib ipaddr cstruct stringext uri vchan tls"
 
 case "$OCAML_VERSION,$OPAM_VERSION" in
 4.00.1,1.1.0) ppa=avsm/ocaml40+opam11 ;;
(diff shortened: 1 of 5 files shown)


samoht:
This does not compile with `tls.0.3.0`. Which version did you use?

talex5:
0.3.0 is too old. It should work with the tls Git master.

samoht:
Ok, so this needs a new release of `ocaml-tls` with https://github.com/mirleft/ocaml-tls/pull/258

hannesm:
@samoht yes, this is true... and will happen soon. to work on xen, I believe gmp and zarith from mirage-dev need to be upstreamed as well

samoht:
I'm trying to make that PR to compile in #46. Given #25, I'll also try to use the new dynamic flows to simplify the API: we really cannot extend the library as we do currently.

PR #41, opened 2015-01-26
Updating use of Tls.Config.server to match new interface of ocaml-tls

heidi-ann:


avsm:
Dup of 83db584b, so already merged.

PR #45, opened 2015-03-04
Remove the resolver overwrite

samoht:
Related to #6

I'm not sure that's a good idea to merge as sometimes overwriting the resolver is useful but I'm not sure how to fix this properly ... but that patch fixes my problem (https://github.com/mirage/irmin/issues/107)


samoht:
Superseded by #49

PR #46, opened 2015-03-20
Tls

samoht:


samoht:
> `packages/ocaml-planet/ocaml-planet.0.1.1/descr`

The current policy is to remove any `ocaml-` prefix from package names as the repository only contains `ocaml` project.

samoht:
Merging as-is, I'll improve the API later.

Issue #7, opened 2014-06-01
SSL Client Authentication


paurkedal:
I would like to ask for a way to supply a certificate and key for the cohttp client (lwt), as I use SSL client authentication.  I used Cohttp_lwt_unix_net.Ssl_client.sslctx up til now, but I can't find an alternative with the latest version.  I can see the SSL code has been moved here, but depending on how it's solved, Cohttp may need adjustments too.

The simplest fix may be to re-expose the SSL context as a global.

An alternative to avoid modifying globals after program initialization:
1. Add an initialization function which picks up defaults for the global context from environment variables (e.g. CONDUIT_C_CERT, CONDUIT_C_CERTKEY, CONDUIT_S_CERT, CONDUIT_S_CERTKEY, CONDUIT_S_CACERT_FILE, ...).  This can be implemented by client code if the context is exposed, which might be preferable for applications complex enough to have it's own configuration file and environment.
2. Add parameters to override these.  I could be a single record of SSL parameters passed as ~tls:{tls_defaults with ...}.  If these/this parameter is present, create a new context, otherwise use the global one.

But the simple fix suffice for my use.

samoht:
Now that we have server-side TLS, would be nice to also add client-side TLS in the 0.8.0 release.

darioteixeira:
Is there an update on this? I have a situation where the TLS client needs to authenticate itself to the server, and I was hoping to continue using the main branch of Conduit...

hannesm:
@darioteixeira #223 seems to implement client authentication... but as said in the review there, I don't understand `conduit` (and haven't tested the PR)

dinosaure:
According to #303, this issue should be close. Then, the new API of `conduit` lets the user to define its own TLS/SSL configuration needed to start a client connection such as:
```ocaml
let my_tls_config = Tls.Config.client ~authenticator ()

let my_resolver domain-name =
  Conduit_lwt.TCP.resolve ~port:443 domain_name >>= function
  | Some edn -> Lwt.return_some (edn, my_tls_config)
  | None -> Lwt.return_none

let resolvers =
  Conduit.empty
  |> Conduit.add Conduit_lwt_tls.TCP.protocol my_resolver

let () =
  Conduit.resolve ~resolvers my_website >>= fun flow ->
  ...
```

paurkedal:
Nice, I can see the new API along with the tls and x509 libraries cover authenticating the client, though it seems too early to test it with cohttp.

[...]

Issue #42, opened 2015-02-05
Binding on Unix to particular IP not working


avsm:
Using `cohttp-server-lwt` it appears to bind to all interfaces even with a specific `-s`

samoht:
Seems that `cohttp-server-lwt` doesn't call the conduit init function -- actually it does nothing at all with the `host` parameter:

https://github.com/avsm/ocaml-cohttp/blob/master/bin/cohttp_server_lwt.ml#L92

samoht:
I've moved the issue in cohttp's bugtracker.

PR #47, opened 2015-03-20
prepare the release

samoht:
Would be great to have fixes for #7 and #45 before the actual release


avsm:
#42 is also suspicious

samoht:
I've just rebased the release runes as the rest has already been merged to master.

PR #54, opened 2015-03-25
Fix for leaked pipes

### LINKED ISSUE #53: Conduit_async_ssl. ssl_listen Leaks Pipes
See https://github.com/mirage/ocaml-cohttp/issues/287 

It appears that Conduit_async_ssl.ssl_listen creates a couple of pipes that are never closed. See the above ticket for reproduction of this bug.

trevorsummerssmith:
NB I am not familiar with this codebase so not totally clear if this
is the correct fix for this issue

Fixes #53 for me

### DIFF
--- a/lib/conduit_async_ssl.ml
+++ b/lib/conduit_async_ssl.ml
@@ -50,4 +50,9 @@ let ssl_listen ~crt_file ~key_file rd wr =
   >>= fun app_rd ->
   Writer.of_pipe (Info.of_string "async_conduit_ssl_writer") app_wr
   >>| fun (app_wr,_) ->
+  (* Close the pipes we created for ssl when we're done! *)
+  don't_wait_for (
+    Pipe.closed net_to_ssl >>= fun () ->
+    Deferred.all_ignore [Writer.close app_wr; Reader.close app_rd]
+  );
   app_rd, app_wr


lib/conduit_async_ssl.ml
```diff
@@ -50,4 +50,9 @@ let ssl_listen ~crt_file ~key_file rd wr =
   >>= fun app_rd ->
   Writer.of_pipe (Info.of_string "async_conduit_ssl_writer") app_wr
   >>| fun (app_wr,_) ->
+  (* Close the pipes we created for ssl when we're done! *)
+  don't_wait_for (
+    Pipe.closed net_to_ssl >>= fun () ->
+    Deferred.all_ignore [Reader.close app_rd; Writer.close app_wr]
```
rgrinberg: We usually specify the order in cases like this. So let's close the writer first.

trevorsummerssmith:
@rgrinberg I switched the order as per your suggestion. I amended the original commit.

rgrinberg:
+1. I actually thought `all_ignore` left the order unspecified but it's left to right. I.e. basically `sequence_`.

lib/conduit_async_ssl.ml
```diff
@@ -50,4 +50,9 @@ let ssl_listen ~crt_file ~key_file rd wr =
   >>= fun app_rd ->
```
rgrinberg: Not being that familiar with the code of conduit I'm wondering if it's possible for `net_to_ssl` to be closed before your cleanup call is made. E.g. after the `app_rd` is created but before `app_wr`. Admittedly, that's a little contrived.

trevorsummerssmith: @rgrinberg I don't know either. I am eager to hear from a core committer about this.

avsm: As the core committer, I'm not much more help -- the Async_ssl interface is terribly documented.  Am reading the source code now...

avsm: I've tried your patch for 10 minutes with a curl running against cohttp lib_test/test_net_async_server.native, and there's no observable memory leak from a lot of https:// connections with this patch. So merging this...

PR #64, opened 2015-04-27
Support the new entropy mechanism for mirage/tls

samoht:


samoht:
This is part of https://github.com/mirage/mirage-dev/pull/52 and  https://github.com/mirleft/ocaml-nocrypto/issues/55

lib/conduit_lwt_tls.ml
```diff
@@ -17,7 +17,7 @@
 
 open Lwt
 
-let _ = Tls_lwt.rng_init ()
+let _ = Nocrypto_entropy_lwt.initialize ()
```
hannesm: and we should do this properly, actually... meaning within `lwt_main` loop... not via this racy piece here (yeah, my fault)

samoht:
I'm merging this one for now. @hannesm please feel free to send a better fix!

lib/conduit_lwt_tls.ml
avsm: I created #67 to track this comment.

Issue #85, opened 2015-07-24
Documentation misses some doc-comments


smondet:
The comments on the variants (like the difference between `` `TLS``, `` `TLS_native``, etc.) are not propagated there:
http://mirage.github.io/ocaml-conduit/Conduit_lwt_unix.html#TYPEclient
(but they are there: https://github.com/mirage/ocaml-conduit/blob/master/lib/conduit_lwt_unix.mli#L34)

dbuenzli:
`ocamldoc` has never supported the comments on variant cases.

dsheets:
The issue with attaching documentation to polymorphic variant constructors is that polymorphic variant constructors are part of type *expressions* whereas normal variant constructors are part of type *definitions*. 

That is, writing ```type t = [ `Foo | `Bar ]``` is basically like writing ```type t = int * string``` and may show up in unascribed cases as, e.g.,  ```[ `Foo | `Bar ] -> 'a```. codoc will also not support documentation comments on polymorphic variant constructors until this distinction is overcome in the ocamldoc parser.

dbuenzli:
> codoc will also not support documentation

Ach shade. I thought `codoc` would overcome this.

dsheets:
Considering this further, I believe it will require compiler modification to attach doc nodes to subcomponents of type expressions. I know of at least 3 groups that want this feature now so it's probably a good time to file a mantis issue or raise it on a mailing list.

dbuenzli:
> good time to file a mantis issue

There's already a long standing one http://caml.inria.fr/mantis/view.php?id=4347

samoht:
Online doc updated: http://mirage.github.io/ocaml-conduit/Conduit_lwt_unix.html#TYPEclient

PR #86, opened 2015-07-26
Bypass unsupported ocamldoc for polymorphic variants

smondet:
Fixing #85, the old school way.

screenshot 2015-07-26 11 20 46


samoht:
Thanks!

PR #95, opened 2015-09-15
Prepare the release

samoht:


CHANGES
```diff
@@ -1,3 +1,12 @@
+0.8.8 (2015-09-15)
+* Expoe a new functor `Conduit_mirage.With_tcp` (#92, by @Drup)
```
avsm: Expoe/Expose

CHANGES
```diff
@@ -1,3 +1,12 @@
...
+* Expose a new functor: `Resolver_mirage.Make_with_stack` (#92, by @Drup)
+* Expose `Resolver_mirage.S`, the signature for Mirage's conduit resolvers.
+  Resolvers now also expose their `DNS` implmentation as a submodule of
+  `Resolver_mirage.S` (#92, by @Drup)
```
avsm: hm, but what if the resolver doesnt support DNS?

samoht: That description is indeed not very clear, I'll change it. The functors we expose in `Resolver_mirage` all convert a DNS resolver into a conduit resolver. If you don't want DNS, don't use `Resolver_mirage` I guess. I'm not sure the current combinator in the DSL will work very well with that...

avsm: thanks, makes sense

PR #96, opened 2015-10-10
Support servers being started on-demand by launchd

djs55:
In on-demand mode, launchd reads the .plist file and listens on the
specified Unix domain sockets / TCP ports / other things. When something
interesting happens, launchd will spawn the service and expect it to
call `launch_activate_sockets` to receive the listening sockets. The
client is expected to call `accept` on these sockets.

This is very similar to the existing socket server, except that launchd
is the one who creates and listens on the sockets.

Signed-off-by: David Scott <dave.scott@unikernel.com>

### DIFF
--- a/META.in
+++ b/META.in
@@ -21,7 +21,7 @@ package "lwt" (
 package "lwt-unix" (
  version = "@VERSION@"
  description = "Resolve URIs into Lwt channels under Unix"
- requires = "conduit.lwt @LWT_UNIX_REQUIRES@ @VCHAN_LWT_REQUIRES@"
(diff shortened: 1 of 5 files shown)


avsm:
And so the depopts list for launchd continues to grow. Gulp :)

avsm:
Could you please move djs55/ocaml-launchd into the mirage org before importing this?

djs55:
Move completed. It's kinda nice that conduit hides all this depopt stuff
from my app, so I'm not complaining!

On Mon, Oct 12, 2015 at 2:38 PM, Anil Madhavapeddy <notifications@github.com
> wrote:

> Could you please move djs55/ocaml-launchd into the mirage org before
> importing this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mirage/ocaml-conduit/pull/96#issuecomment-147402561>.
>



-- 
Dave Scott

PR #101, opened 2015-12-12
lwt/ocaml-tls server fix

### LINKED ISSUE #97: Different behavior between nqsb-tls and openssl
Maybe I'm missing something obivous, but with `CONDUIT_TLS=native` the HTTP requests seems to be handled sequentially.

This is my server:

```ocaml
#use "topfind";;
#require "cohttp.lwt";;
let say fmt = Printf.(ksprintf (printf "%s\n%!") fmt)

let mode =
  let certfile = "_test_env/test-cert.pem" in
  let keyfile =  "_test_env/test-key.pem" in
  let port = 4242 in
  `TLS (
    (* `TLS means that conduit will do:
       match Sys.getenv "CONDUIT_TLS" with
       | "native" | "Native" | "NATIVE" -> Native
       | _ -> OpenSSL *)
    `Crt_file_path certfile,
    `Key_file_path keyfile,
    `No_password, `Port port)
open Lwt
let req_callback (_, conn_id) request body =
  let id = Cohttp.Connection.to_string conn_id in
  say "Connection %s, request %s" id
    (Cohttp.Request.sexp_of_t request |> Sexplib.Sexp.to_string);
  match Uri.path (Cohttp.Request.uri request) with
  | "/sleep" ->
    Lwt_unix.sleep 10.
    >>= fun () ->
    Cohttp_lwt_unix.Server.respond_string ~status:`OK  ~body:"slept" ()
  | other ->
    Cohttp_lwt_unix.Server.respond_string ~status:`OK  ~body:"hello" ()

let () =
  Lwt_main.run
    Cohttp_lwt_unix.Server.(
      create ~mode (make ~callback:req_callback ()))
```

### OpenSSL

If I run ` CONDUIT_TLS=openssl ocaml server.ml`

Then I can run `curl -k https://127.0.0.1:4242/sleep` on one terminal, and many times `curl -k https://127.0.0.1:4242/hello` on another terminal, everything is fine.

### OCaml-TLS

When running `CONDUIT_TLS=native ocaml server.ml`, if I have  `curl -k https://127.0.0.1:4242/sleep` on one shell, any call to  `curl -k https://127.0.0.1:4242/hello` on another terminal will block until the `/sleep` is done.

rgrinberg:
Server in `Conduit_lwt_tls` waits for a user callback to finish before
accepting more connections. Instead, it should only wait until the
connection is accepted and detach the client callback

Fix #97

### DIFF
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -19,6 +19,14 @@ open Lwt
 
 let _ = Nocrypto_entropy_lwt.initialize ()
 
+let safe_close t =
+  Lwt.catch
+    (fun () -> Lwt_io.close t)
+    (fun _ -> return_unit)
+
+let close (ic, oc) =
+  Lwt.join [ safe_close oc; safe_close ic ]
+
 let with_socket sockaddr f =
   let fd = Lwt_unix.socket (Unix.domain_of_sockaddr sockaddr) Unix.SOCK_STREAM 0 in
   Lwt.catch (fun () -> f fd) (fun e ->
@@ -65,7 +73,7 @@ module Server = struct
     let events = match timeout with
       | None -> [c]
       | Some t -> [c; (Lwt_unix.sleep (float_of_int t)) ] in
-    Lwt.ignore_result (Lwt.pick events)
+    Lwt.ignore_result (Lwt.pick events >>= fun () -> close (ic, oc))
 
   let init ?(nconn=20) ~certfile ~keyfile
         ?(stop = fst (Lwt.wait ())) ?timeout sa callback =
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -65,7 +65,7 @@ module Server = struct
     let events = match timeout with
       | None -> [c]
       | Some t -> [c; (Lwt_unix.sleep (float_of_int t)) ] in
-    Lwt.pick events
+    Lwt.ignore_result (Lwt.pick events)
 
   let init ?(nconn=20) ~certfile ~keyfile
         ?(stop = fst (Lwt.wait ())) ?timeout sa callback =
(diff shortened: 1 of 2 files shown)


rgrinberg:
@samoht @avsm could one of you review and/or merge?

dsheets:
My only comment is that `ignore_result` is slightly more confusing than `async` but it does avoid a closure allocation. Otherwise, I think this looks good.

hannesm:
I wonder why not preserving the same shape in ocaml-tls as is in ssl, namely doing the ignore_result (or `let _`) in `process_accept`?

dsheets:
I agree with Hannes that the implementations should be parallel but the `let _` in the ssl version scares me.

rgrinberg:
Sure, I also prefer to `Lwt.ignore_result` over `let _`.

rgrinberg:
@hannesm @dsheets I've changed both to use `Lwt.ignore_result` and moved it to `process_accept`.

hannesm:
great, looks good to me! looking at the difference between ssl and tls, it seems to me that the tls part is leaking file descriptors since nobody calls close after executing the callback... must've been a sloppy programmer (me! ;)) who wrote that piece of code... sorry

rgrinberg:
Good eye, we should really just unify these accept loops though. I wonder why `Lwt_io.establish_server` wasn't used btw.

rgrinberg:
@hannesm I fixed the FD leak as well.

rgrinberg:
@avsm @samoht ping.

avsm:
@rgrinberg apologies for delay; fix looks good to me! Code duplication can be tackled in another PR.

avsm:
We could use `Lwt_io.establish_server` here -- I'm just used to writing that code out so that sockopts can be added as well more easily (`NODELAY` and such).

PR #104, opened 2015-12-20
close socket when ssl_accept fails

edwintorok:
When `Lwt_ssl.ssl_accept` fails conduit doesn't close the socket which leaves these clients hanging until a timeout occurs. (Try running `cipherscan` for example)
When SSL handshake fails (e.g. due to ciphersuite/protocol version mismatch) the socket should be closed.

Would be nice if we could also log the failure, but I haven't seen `Lwt_log` or `logs` used anywhere in conduit/cohttp yet, so I just fixed the socket close for now.

### DIFF
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -53,9 +53,12 @@ module Server = struct
 
   let accept config s =
     Lwt_unix.accept s >>= fun (fd, sa) ->
-    Tls_lwt.Unix.server_of_fd config fd >|= fun t ->
-    let ic, oc = Tls_lwt.of_t t in
-    (fd, ic, oc)
+    Lwt.try_bind (fun () ->
+        Tls_lwt.Unix.server_of_fd config fd)
+      (fun t ->
+        let ic, oc = Tls_lwt.of_t t in
+        return (fd, ic, oc))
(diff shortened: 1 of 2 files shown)


rgrinberg:
Looks good to me. Seems like the TLS backend suffers from this this as well however:

https://github.com/rgrinberg/ocaml-conduit/blob/b5a8a55361f08d4a882919ec794dbd7f18e423eb/lib/conduit_lwt_unix_ssl.ml#L63-L66

What do you think? Should we apply this patch there as well.

edwintorok:
Yes looks like a similar patch is needed here too: https://github.com/rgrinberg/ocaml-conduit/blob/b5a8a55361f08d4a882919ec794dbd7f18e423eb/lib/conduit_lwt_tls.ml#L56
And possibly in `conduit_async.ml` too?

rgrinberg:
Yes, that's definitely another instance. Good catch.

Likely not for async because we don't have our own accept loop there. It's provided by async_ssl.

edwintorok:
Hmm would be better to have some tests for this, but should those tests go into Cohttp? I only see async and mirage tests in conduit itself.

rgrinberg:
I don't think the async tests in conduit are automatic anyway so I would definitely accept some proper tests in cohttp if that's easier for you.

rgrinberg:
Looks good. I just realized that my original link was bad. Sorry!

PR #105, opened 2015-12-25
0.10.0 release

avsm:
0.10.0 (2015-12-25):
* Add support for CA certificates in [Conduit_async.serve] (#98).
* Fix file descriptor leak in Lwt backend (#101 from @hannesm).
* Server in `Conduit_lwt_tls` waits for a user callback to finish
  before accepting more connections. Instead, it should only wait
  until the connection is accepted and detach client callback (#97).
* Close socket when `ssl_accept` fails, e.g. when cipher negotiation
  mismatch (#104).


smondet:
Hi ! Why wasn't `0.10.0` completely released and pushed to the opam-repo?

rgrinberg:
Looks like it wasn't tagged either. I assume @avsm just forgot, I'll pick up the slack.

PR #111, opened 2016-01-25
Fixed server stop

### LINKED ISSUE #110: Can't properly stop lwt server
There appear to be two ways to stop a server started with `Conduit_lwt_unix.serve`, and neither seems to work correctly.
- If I pass `~stop` and wake up the stop thread, the server will accept the next connection and then exit its loop. However, the server thread never returns and the listening socket remains open.
- If I `Lwt.cancel` the server thread, the server immediately stops accepting connections and closes all existing connections. The server thread returns, but the listening socket remains open.

In both cases, I'm left with a listening socket that never accepts connections and prevents me from starting a new server on that port.

I think the correct behaviour in both cases is to close the listening socket when the accept loop exits, but I'm not sure why there are two mechanisms for this or what the intended differences are.

prepor:
It should fix #110, ie:
- Properly stops thread with server
- Eliminate fd leaks on server stop
- Does not create never-terminated tasks

### DIFF
--- a/lib/conduit_lwt_unix.ml
+++ b/lib/conduit_lwt_unix.ml
@@ -225,19 +225,19 @@ module Sockaddr_server = struct
     let s = match on with
     | `Socket s -> s
     | `Sockaddr sockaddr -> init_socket sockaddr in
-    async (fun () ->
-      stop >>= fun () ->
-      cont := false;
-      return_unit
-    );
+    let stop' = Lwt.map (fun () -> `Stop) stop in
     let rec loop () =
-      if not !cont then return_unit
-      else
-        Lwt_unix.accept s >>=
-        process_accept ?timeout callback >>= fun () ->
+      let accept = Lwt_unix.accept s in
+      Lwt.choose [Lwt.map (fun v -> `Accept v) accept;
+                  stop'] >>= function
+      | `Stop ->
+        Lwt.cancel accept;
+        Lwt.return_unit
+      | `Accept v ->
+        process_accept ?timeout callback v >>= fun () ->
         loop ()
     in
-    loop ()
+    Lwt.finalize loop (fun () -> Lwt_unix.close s)
 end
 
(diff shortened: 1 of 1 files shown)


jerith:
Thanks for this!

I'm sadly not qualified to review it properly, but I'll definitely try it out later in the week to see if it fixes my problem.

prepor:
Actually there is memory leak :) See https://github.com/ocsigen/lwt/issues/211
I'll rewrite it some other way.

prepor:
I fixed memory leak

lib/conduit_lwt_unix.ml
```diff
@@ -225,19 +225,19 @@ module Sockaddr_server = struct
...
+                  stop'] >>= function
+      | `Stop ->
+        Lwt.cancel accept;
+        Lwt.return_unit
+      | `Accept v ->
+        process_accept ?timeout callback v >>= fun () ->
         loop ()
     in
-    loop ()
+    Lwt.finalize loop (fun () -> Lwt_unix.close s)
```
avsm: It would be better to explicitly close this socket in the `loop` -- why not call it in the `Stop` pattern match above rather than wait for the GC?

prepor: Lwt.finalize is not about GC. It's just run `f` regardless of success or fail. If Lwt_unix.accept will fail for any reason `Stop code will not run.

rgrinberg:
@avsm I think this is good. Do you have any objections to this?

@prepor Do you mind applying this to all lwt backends until we fix #112

prepor:
@rgrinberg i'm sorry, but I currently not interested in this code, as I don't use lwt version of cohttp / conduit.

avsm:
no objections to this, but would be really good to add a compose.yml test to test out server stop. I believe the Travis Docker runs compose.yml

vbmithr:
Thanks!

Issue #113, opened 2016-02-02
Conduit_lwt_unix_ssl leaks file descriptors


edwintorok:
`Lwt_io.close` on the channel obtained from` Lwt_ssl.out_channel_of_descr` leaks the underlying FD if `ssl_shutdown` fails. This can occur if the client has disconnected uncleanly (e.g. TCP  RST, or simply gone in the case of a mobile client, etc.).
As a consequence my cohttp.lwt server died every few weeks due to running out of file descriptors.

Pull request coming up

Issue #109, opened 2016-01-13
limit maximum number of active sockets


edwintorok:
When Conduit runs out of sockets in `Conduit_lwt_unix_ssl` it loops consuming 100% CPU due to ignoring all exceptions: https://github.com/mirage/ocaml-conduit/blob/master/lib/conduit_lwt_unix_ssl.ml#L103

```
accept(3, 0x7ffc3bd61d50, [112])        = -1 EMFILE (Too many open files)
accept(3, 0x7ffc3bd61d50, [112])        = -1 EMFILE (Too many open files)
accept(3, 0x7ffc3bd61d50, [112])        = -1 EMFILE (Too many open files)
accept(3, 0x7ffc3bd61d50, [112])        = -1 EMFILE (Too many open files)
```

The behaviour is different from `Conduit_lwt_unix` which would apparently just quit the accept loop.

It would be better if Conduit would count the number of globally accepted connections and throttle (wait) until the number of connections drops below a maximum.
FWIW Ocsigen uses a maximum number of connections =  350 by default.

Since this would be a global setting (if you have multiple conduits listening on different ports they should use the same shared counter) would it make sense to have the counter in `Conduit.mli` and have all conduit implementions use that?
Obviously throttling itself has to be implemented differently in Lwt and Async.
For now I've added an ugly hack on my branch, but this maximum should also be configurable.

Issue #102, opened 2015-12-16
support TLSv1.1 and TLSv1.2 with openssl backend


edwintorok:
Currently ocaml-conduit explicitly requests only TLSv1, would be nice if it supported TLSv1.1, v1.2 and whatever new protocols there will be.
To do that you apparently have to request SSLv23 and explicitly disable SSLv2 and SSLv3 then.

edwintorok:
This allows cohttp_curl_lwt to connect to TLSv1.2-only servers, and cohttp_server_lwt to accept TLSv1.2-only clients: https://github.com/mirage/ocaml-conduit/commit/bf3619ae0b13353aa6ff31eb541e8fc3e5d076f1
I haven't tested that SSLv3 is properly disabled because AFAIK Debian disables SSLv3 at build time.
I haven't submitted a pull request yet because I want to add a test too: TLSv1.2 server, default client; and TLSv1.2 client and default server.

PR #115, opened 2016-02-02
Ssl leak / EMFILE / TLSv1.2 fixes

edwintorok:
Small self-contained fixes for #113, #102  and workaround for #109, and a test that can be run with `tests/unix/run.sh`.
It is not integrated with Travis, not sure how to do that or if it should be integrated at all.

### DIFF
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -25,7 +25,8 @@ let safe_close t =
     (fun _ -> return_unit)
 
 let close (ic, oc) =
-  Lwt.join [ safe_close oc; safe_close ic ]
+  safe_close oc >>= fun () ->
+  safe_close ic
 
 let with_socket sockaddr f =
   let fd = Lwt_unix.socket (Unix.domain_of_sockaddr sockaddr) Unix.SOCK_STREAM 0 in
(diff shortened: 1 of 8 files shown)


edwintorok:
This has previously failed to build due to async, I've rebased it on master now.

rgrinberg:
Looks good and congrats on the first test for this repo. A couple of questions.

1) Do any of these fixes need to be backported to the other lwt backends? You know our current situation with the lwt based servers...

2) Do we have to expose `default_ctx` in the mli? Seems like this isn't totally required.

edwintorok:
Thanks for the review.
1)  the non-mirage lwt backends should be all fine: I updated both OpenSSL/tls backends, and the `yield` fix is not needed for non-ssl (it doesn't infloop, it just exits on accept failure). TBH I haven't tried this test on the mirage backends yet,but conduit just calls the STACKV4 implementation, it doesn't appear that anything in `conduit_mirage.ml` could leak.
2) I'll drop it, and we can figure something better for the next release. I'd prefer if we have just an interface where you specify min/max TLS versions and cipherlist that I can use with both OpenSSL and ocaml-tls backends. (sometimes it is necessary to tweak these when debugging connectivity problems with various clients)

Issue #114, opened 2016-02-02
Conduit_lwt_{ssl/tls} serializes on Ssl accept


edwintorok:
When connecting to a remote server the SSL handshake can take e.g. 150ms, and due to the code below and no new connections were accepted until previous one finished the SSL handshake:
`accept config s >|= process_accept ~timeout callback`

It'd be better to perform the accept and handle the SSL handshake asynchronously like the callback itself. This can be implemented nicely by unifying the accept logic into `Conduit_lwt_unix` as in #112.

dinosaure:
It's hard to say that the new version of `conduit` fix this issue. From what I know, the issue should not be fixed when we still continue to use `Lwt_ssl.accept` when you want to use `Conduit.Service.accept` (see [this lines](https://github.com/mirage/ocaml-conduit/blob/66fd34a9dc1dc2507f6c20752e589ae53cc4ccf8/src/lwt-ssl/conduit_lwt_ssl.ml#L100-L107)). However, I think it will be hard to do an asynchronous handshake (as `conduit-lwt-tls` does).

From the API provided by `lwt_ssl`, we should be able to use `ssl_accept_handshake` and keep an internal state which informs to us if the current used socket is _initialized_ or not but it will complexify a bit the implementation.

If you want such behavior (delay the handshake to `Conduit.{recv,send}`), feel free to open an issue and I will happy to investigate a possible solution about that. I will close this issue. Thanks.

Issue #112, opened 2016-01-26
unify accept loop and process accept in lwt (conduit_lwt_unix,conduit_lwt_tls, conduit_lwt_unix_ssl)


edwintorok:
bugfixes/changes would have to be applied 3 times otherwise, e.g. https://github.com/mirage/ocaml-conduit/pull/111
To avoid creating circular dependencies perhaps should be a `conduit_lwt.ml` for shared code.

dinosaure:
From the new version of `conduit`, it seems that the only diff between what you did and the current implementation of the server loop is the use of `Lwt.cancel` on your side. You can see the impl. here:
https://github.com/mirage/ocaml-conduit/blob/66fd34a9dc1dc2507f6c20752e589ae53cc4ccf8/src/lwt/conduit_lwt.ml#L66-L81

Should we re-integrate `Lwt.cancel` @edwintorok?

Issue #107, opened 2015-12-30
channels hang when stream raises exception


edwintorok:
When a body stream raises an exception the connection is kept open (probably forever).

Since the headers were already sent we don't have much choice but to close the channel, with chunked encoding `curl` would report a proper error then:

```
curl: (18) transfer closed with outstanding read data remaining
```

I was testing a fix like this

``` ocaml
-    Lwt.ignore_result (Lwt.pick events >>= fun () -> close (ic,oc))
+    Lwt.finalize (fun () -> Lwt.pick events) (fun () -> close (ic,oc)) |>
+    Lwt.ignore_result
```

Not sure if `Lwt.finalize` or `Lwt.on_termination` would be preferred here (finalize creates an extra closure?).

dinosaure:
I'm not sure that `conduit.3.0.0` properly fix the error when it's mostly about the _tcp implementation_ used (which is not the part of the `conduit` package). If such error still exists, it's about `conduit-lwt` and the sub-module `Conduit_lwt.TCP` which implements the TCP protocol. As far as I can (and as the main developer of this impl.), I did not notice such behavior on my tests - but I did not test this specific issue.

Due to the major release, I will close this issue but if the bug re-appear, feel free to re-open the issue and I will try to investigate the bug more deeply.

mseri:
I think this is still present in the rever to cohttp 4.0

PR #116, opened 2016-02-02
Fix errors limits

edwintorok:
Fix for #107, #109, #112 and updated tests meant for discussion.

This is a fairly large change (especially the unification), I tried to do it in incremental steps, but please tell me how I should proceed, or if you'd like a different approach.
The reason for the unification is that I kept fixing the same bugs over 3 implementations (unix,  unix+ssl, unix+tls) and one bug got fixed just by unifying the implementations (#114).

Depends on https://github.com/mirage/ocaml-conduit/pull/115 and https://github.com/mirage/ocaml-conduit/pull/111, so please don't merge yet. Once those 2 other PRs are decided I'll rebase this one.


lib/conduit_lwt_unix_ssl.ml
```diff
@@ -16,34 +16,23 @@
...
-  Lwt.catch (fun () -> f fd) (fun e ->
-      Lwt.catch (fun () -> Lwt_unix.close fd) (fun _ -> return_unit) >>= fun () ->
-      fail e
-    )
-
 module Client = struct
   (* SSL TCP connection *)
-  let t = Ssl.create_context Ssl.TLSv1 Ssl.Client_context
+  let default_ctx = Ssl.create_context Ssl.SSLv23 Ssl.Client_context
+  let () = Ssl.disable_protocols default_ctx [Ssl.SSLv23]
```
avsm: why this change from `Ssl.TLSv1` to `Ssl.SSLv23` followed by the disabling of the older protocol?

avsm:
This is absolutely the right direction, thanks very much! Will review shortly...

lib/conduit_lwt_unix_ssl.ml
edwintorok: Before OpenSSL 1.1.0 the only way to enable TLS1+ is to use `SSLv23_method` which enables all protocols,
and then selectively disable the protocols you don't want ... ugly, see https://github.com/mirage/ocaml-conduit/issues/102 and https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html.
After OpenSSL 1.1.0 there is a nicer API https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_min_proto_version.html that sets protocol version limits but `ocaml-ssl` doesn't support that yet.

I didn't want to enable just `TLSv1.2` because when TLSv1.3 appears it wouldn't be automatically supported.

vbmithr:
@avsm @rgrinberg 
What's the status of this? I think we should go ahead, review and hopefully merge.
Some bugs have been there for too long…

vbmithr:
Ok, it depends on #111.

vbmithr:
@edwintorok Hi Edwin, sorry for the lag. Could you go ahead on this?

edwintorok:
I'll need to rebase/redo some of the commits, the underlying code has changed quite a lot (in fact some unification already happened https://github.com/mirage/ocaml-conduit/commit/e555367a0a887eda598f97294222edfad5188a0c which is good). I'll update this PR when I'm done.

vbmithr:
On 06/12/2016 11:10, Török Edwin wrote:
> I'll update this PR when I'm done.

Great! Thank you very much.

rgrinberg:
I've just completed most of the unification work in #173. So let's just get the bug fixes from this PR + the good interface changes.

edwintorok:
Thanks @rgrinberg, your unification reduced the number of commits in this PR significantly, I pushed a simplified branch. I have updated the tests (in `tests/unix/cdtest`) and they pass, however I'd like to do some more tests to make sure that all the bugs referenced in this PR are indeed gone.

I'd  also like to run `tests/unix/run.sh` and `tests/unix/run_tls.sh` with Travis, do you have some pointers on what I need to modify in `.travis.yml` to make that happen?

rgrinberg:
Thanks for working on this. Let me know how I can help.

>I'd also like to run `tests/unix/run.sh` and `tests/unix/run_tls.sh` with
>Travis, do you have some pointers on what I need to modify in `.travis.yml` to
>make that happen?

I think you should just create a make target that exercises all the tests and
then add this command to `build-test` in the opam file.

That should build the tests automatically.

rgrinberg:
Hmm, how come lwt.ssl isn't being install on travis?

# ocamlbuild -use-ocamlfind -classic-display -cflags "-w A-4-33-40-41-42-43-34-44" cdtest.native
# ocamlfind ocamldep -package lwt.ssl -package lwt.unix -package tls.lwt -package ipaddr.unix -package sexplib -package ipaddr -package cstruct -package uri -package stringext -package logs -modules tests/unix/cdtest.ml > tests/unix/cdtest.ml.depends
# + ocamlfind ocamldep -package lwt.ssl -package lwt.unix -package tls.lwt -package ipaddr.unix -package sexplib -package ipaddr -package cstruct -package uri -package stringext -package logs -modules tests/unix/cdtest.ml > tests/unix/cdtest.ml.depends
# ocamlfind: Package `lwt.ssl' not found
# Command exited with code 2.
# Makefile:74: recipe for target 'cdtest' failed
### stderr ###
# make: *** [cdtest] Error 10

avsm:
Is it 4.04? No ssl or tls there yet last I looked

rgrinberg:
4.03 is also erroring out like this (and the other builds too)

edwintorok:
Is there a phase in the travis run where the depopts aren't installed yet? (I don't see anything in the build logs about building ssl or async). Anyway I should add some conditionals on that test, will try to run that travis  test script locally to see what happens.

edwintorok:
Added the conditionals, the tests pass, but they are not actually run, looks like the tests are not re-run after installing the depopts.

Also found a problem with the `cdtest_tls` test: it is not deterministic under high CPU load (i.e. `( while true; do  echo -n;done )` on all your cores) it will fail with an error when the client tries to read the message sent by the server:

```
cdtest_tls.native: main: Server running on port 9192
cdtest_tls.native: main: Server: Callback started.
cdtest_tls.native: main: Connected!
cdtest_tls.native: main: Written hello.
cdtest_tls.native: main: Server: read hello.
cdtest_tls.native: main: Got correct msg, disconnecting.
cdtest_tls.native: main: Server: Callback started.
cdtest_tls.native: main: Connected!
cdtest_tls.native: main: Written hello.
cdtest_tls.native: main: Server: read hello.
./cdtest_tls.native: "Tls_lwt.write" failed: Broken pipe
```

Doing a graceful shutdown on the server with `Tls_lwt.of_t ~close` works this around, but doesn't seem right:
```ocaml
  Lwt.catch (fun () -> Tls_lwt.Unix.read t (Cstruct.create 1) >|= ignore)
     (fun _ -> Lwt.return_unit) >>= fun () ->
  Tls_lwt.Unix.close t
```

rgrinberg:
Perhaps we can list the depopts as `{test}` dependencies as well?

edwintorok:
The tests run now, but some fail: debian-stable due to the `Tls_lwt` shutdown problem, centos-6 due to ctypes segfault (reported), fedora-24/alpine due to lack of m4 (probably because I pinned ocaml-ci-scripts to HEAD to run this test).

If you prefer I can remove the `.travis.yml` changes from this PR, and put them into a separate PR and enable it only once all tests pass.

lib/conduit_lwt_server.mli
```diff
@@ -1,6 +1,8 @@
 
 val close : 'a Lwt_io.channel * 'b Lwt_io.channel -> unit Lwt.t
 
+val set_max_active : int -> unit
```
rgrinberg: Is there any way to do this on a per server basis? This doesn't seem like a constant that can be universal for all servers started with conduit.

rgrinberg:
I've cherry picked the first commit and your lwt_tls fix will also be merged shortly. So feel free to rebase after that.

Would be nice to have the CI changes in a separate PR but we're likely not to wait and just merge them straight away. We must read the travis often enough anyway and false negatives are better than false positives.

lib/conduit_lwt_server.mli
edwintorok: It would be better to leave this decision up to the user of conduit, you might want to have a global limit, per server limit or some combination of them. I went with a global limit here because the underlying resource (process file descriptors) are a global resource too (unless you're running Mirage I guess).
How about  an optional callback argument for `val serve` instead, which would get called with the number of active connections on the current server, the number of active connections globally, and decides what to do with the connection  (accept it, or sleep until either of these two changes).

rgrinberg: I see, that makes sense. But although it's a global resource we can only share this limit between conduit servers here.  

Anyway, I say let's leave it as is for now. If we need more flexibility, then we can consider your proposal here. Seems like adding this limit was motivated by a concrete need you had so let's wait until something concrete like that pops up.

rgrinberg:
This PR is ready after rebase.

edwintorok:
Thanks, pushed a rebased branch, lets wait until travis finishes.

edwintorok:
Ok, its got the m4 failure from last time (not the same ones because meantime ocaml-ci-scripts got reverted back to not running the depopts tests until some other problems are fixed). If you are OK with having false negatives in the CI for now, then please go  ahead and merge it.

rgrinberg:
Yeah, this is good enough for now.

PR #117, opened 2016-02-20
Update opam file

rgrinberg:
* ocamlfind is a `{build}` dependency
* ocamlbuild is an explicit dependency

trivial fixes


rgrinberg:
Not sure where travis went but this seems harmless enough.

PR #118, opened 2016-02-21
Multi distro travis

avsm:
This also lays the groundwork for running server/client tests in lib_test and spotting resource leaks, which I'll do next.


avsm:
This one bumps sexplib to compile again, and then the next release will use pa_sexp_conv (or perhaps just ppx)

rgrinberg:
Sorry for the bike shedding but why not `0.10.1`? I don't see any new feature or compatibility breaking here.

PR #122, opened 2016-03-10
Port to ppx

rgrinberg:
First commit explains why we must use ppx_driver over ppx_deriving for now.

The only thing that remains to be done here is to somehow tell ocamlbuild that
conduit_config.mlh is a dependency for every file that is being preprocessed.

Any ocamlbuild gurus can help out with this?

~~gasche samoht Drup dbuenzli agarwal~~ (removed the mentions)

Also, we need to wait for TLS to switch over to ppx as well.

PR #124, opened 2016-03-12
Use pa_sexp_conv and bump to 0.11

avsm:
This lets us leap the Core >=113.24 constraint without a full switch to ppx (which requires some dependent changes in #122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ script: bash ./.travis-docker.sh
 env:
  global:
    - PACKAGE="conduit"
-   - DEPOPTS="async_ssl lwt mirage-dns ssl tls"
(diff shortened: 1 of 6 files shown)


dsheets:
`opam` file says 4.02.0 but `CHANGES` says 4.02.3. Is 4.02.3 really necessary? Is it a ppx thing?

rgrinberg:
Failure on alpine linux:

```
# src/bigstring_stubs.c:36:24: fatal error: sys/endian.h: No such file or directory
```

rgrinberg:
@dsheets Hopefully the next release will not longer use type_conv and ppx_deriving/ppx_type_conv require >= 4.02.3

rgrinberg:
Anyway the alpine issue is a core issue so I'm merging this.

PR #130, opened 2016-06-01
Conduit_async_ssl: use Or_error.t and do not silently ignore errors

vbmithr:


lib/conduit_async.ml
```diff
@@ -77,7 +77,9 @@ let connect ?interrupt dst =
 #if HAVE_ASYNC_SSL
       Tcp.connect ?interrupt (Tcp.to_host_and_port (Ipaddr.to_string ip) port)
       >>= fun (_, rd, wr) ->
-      Conduit_async_ssl.ssl_connect rd wr
+      Conduit_async_ssl.ssl_connect rd wr >>| function
+      | Error err -> failwith (Error.to_string_hum err)
+      | Ok res -> res
```
rgrinberg: `Or_error.ok_exn` please

lib/conduit_async.ml
```diff
@@ -86,9 +88,11 @@ let connect ?interrupt dst =
 #if HAVE_ASYNC_SSL
       Tcp.connect ?interrupt (Tcp.to_host_and_port (Ipaddr.to_string ip) port)
       >>= fun (_, rd, wr) ->
-      let open Ssl in
-      match config with | {version; name; ca_file; ca_path; session; verify} ->
-      Conduit_async_ssl.ssl_connect ?version ?name ?ca_file ?ca_path ?session ?verify rd wr
+      match config with {Ssl.version; name; ca_file; ca_path; session; verify} ->
+        Conduit_async_ssl.ssl_connect ?version ?name ?ca_file ?ca_path ?session ?verify rd wr >>| function
+        | Error err -> failwith (Error.to_string_hum err)
+        | Ok res -> res
```
rgrinberg: ditto

lib/conduit_async.ml
```diff
@@ -146,8 +150,9 @@ let serve
             in
             (crt, key, ca_file, ca_path)
         in
-        Conduit_async_ssl.ssl_listen ?ca_file ?ca_path ~crt_file ~key_file rd wr
-        >>= fun (rd,wr) -> handle_request sock rd wr
+        Conduit_async_ssl.ssl_listen ?ca_file ?ca_path ~crt_file ~key_file rd wr >>= function
+        | Error err -> failwith (Error.to_string_hum err)
```
rgrinberg: `Error.raise err`

lib/conduit_async_ssl.ml
```diff
@@ -26,29 +26,39 @@ let ssl_connect ?version ?name ?ca_file ?ca_path ?session ?verify
...
+    ~app_to_ssl
+    ~ssl_to_app
+    ~net_to_ssl
+    ~ssl_to_net
+    () >>= function
+  | Error err -> return @@ Error err
+  | Ok conn ->
+    verify_connection conn >>= function
+    | false ->
+      return (Result.fail (Error.of_string "Connection verification failed."))
```
rgrinberg: `Or_error.error_string` is simpler IMO but it's a matter of taste

lib/conduit_async_ssl.mli
```diff
@@ -33,7 +33,7 @@ val ssl_connect :
   ?verify:(Ssl.Connection.t -> bool Deferred.t) ->
   Reader.t ->
   Writer.t ->
-  (Reader.t * Writer.t) Deferred.t
+  (Reader.t * Writer.t) Deferred.Or_error.t
```
rgrinberg: One thing that the `Or_error` contract is supposed to guarantee is that these functions shouldn't raise. Are we certain this isn't the case?

vbmithr: Well, I'd say they shouldn't under "reasonable" assumptions.
Pipe.create () can fail if there is no fd left for example, but if it's the case you have bigger problems :)

rgrinberg: Are you sure about `Pipe.create`? I'm pretty sure pipes aren't tied to any particular fd's. It's readers/writers that are fd wrappers.

But anyway if you're reasonably certain then I support this change.

vbmithr: On 01/06/2016 19:20, Rudi Grinberg wrote:
> Are you sure about |Pipe.create|? I'm pretty sure pipes aren't tied to
> any particular fd's. It's readers/writers that are fd wrappers.

Yeah you're right, they're not tied to fds, my bad. So this is even
safer that what I said :)

> But anyway if you're reasonably certain then I support this change.

Yeah, reasonably certain… :) But in the end I agree that it's tricky to
be sure to not breach this contract, I just wanted to do something
similar as Ssl.client that returns an Or_error.t.

The |verify| function could raise, for example. And indeed the
|Reader.of_pipe| can raise if there is no fd left.

I'm not certain anymore what's the best thing to do. I think best would
be wrap everything in a big try_with or do not return Or_error.t types.
What do you think?

rgrinberg: > The |verify| function could raise, for example. And indeed the
> |Reader.of_pipe| can raise if there is no fd left.

About verify, this is something that we should definitely reflect in the
docs. Either that it shouldn't raise or that we'll catch whatever it
raises and return it as Or_error.

> 
> I'm not certain anymore what's the best thing to do. I think best would
> be wrap everything in a big try_with or do not return Or_error.t types.
> What do you think?

My biggest concern is now actually backwards compatibility. But it seems
like you might be the only user :)

https://github.com/search?q=conduit_async_ssl&type=Code&utf8=%E2%9C%93

Out of curiousity, is it possible to just use Conduit_async for your use
case?

Anyway, let's see what the other maintainers say about this interface
change.

@avsm @samoht @dsheets

I will cherry pick the error swallowing fix regardless of course.

vbmithr: On 01/06/2016 21:45, Rudi Grinberg wrote:
> Out of curiousity, is it possible to just use Conduit_async for your use
> case?

Absolutely, I actually changed my code to use |Conduit_async| :)

vbmithr:
Ok, new PR, I changed my mind.

- No interface change anymore in cohttp_async_ssl(just fixing the issue in conduit_async_ssl)
- Interface change in conduit_async: use host string instead of Ipaddr.t that was systematically stringified anyway.
- I also integrate seliopou PR.

Please review and hopefully merge, I really use/need this code.

rgrinberg:
It's a bit hard or me to merge stuff when travis is busted:

let's wait for https://github.com/ocaml/opam-repository/pull/6699 to go in so we can do a proper run.

PR #132, opened 2016-06-07
async: add backlog argument to serve

seliopou:
The backlog optional argument replaced the max_pending_connections argument.


rgrinberg:
Seems like the CI failure is ctypes/async_ssl related. Will merge after taking another look later.

PR #133, opened 2016-06-12
Redoing PR for Travis

rgrinberg:
@vbmithr

### DIFF
--- a/lib/conduit_async.ml
+++ b/lib/conduit_async.ml
@@ -61,30 +61,30 @@ type ic = Reader.t
 type oc = Writer.t
 
 type addr = [
-  | `OpenSSL of string * Ipaddr.t * int
-  | `OpenSSL_with_config of string * Ipaddr.t * int * Ssl.config
-  | `TCP of Ipaddr.t * int
+  | `OpenSSL of string * int
+  | `OpenSSL_with_config of string * int * Ssl.config
+  | `TCP of string * int
   | `Unix_domain_socket of string
 ] [@@deriving sexp]
 
 let connect ?interrupt dst =
   match dst with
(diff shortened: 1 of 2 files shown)


rgrinberg:
Why do we have the ipaddr to string commit? Seems like it's breaking compatibility.

vbmithr:
The ability to use hostnames (that are resolved by Async) was negated by the old interface. The Ipaddr.t provided was anyway converted as a string. In practice, if you want to connect to an URL in text form, how do you get an IP address from it with Ipaddr.t with Async?

This is indeed breaking API change but I think this is better. What do you think? If you don't, do you know any straightforward way to get an Ipaddr.t from a hostname (I'm not using Lwt) ?

rgrinberg:
I see. That's indeed annoying. What about introducing a new set of
constructors that allow people to specify the host names directly?

Since this is a breaking change, I'd like to hear what other maintainers
think.

@dsheets @avsm @samoht

rgrinberg:
@vbmithr OK I've thought over this more and without more feedback I'm in favor of this change.

A couple of things still bug me however:
1) Ipaddr.t isn't used in any of the constructors. I suppose that's alright.
2) Our interface still isn't exposing the "async way" of constructing an endpoint which is the `Tcp.where_to_connect` type. Does it make sense to expose that in another constructor as well perhaps?

vbmithr:
@rgrinberg Yes definitely we need the same constructor as in Async. I would even remove the existing one to leave only this one. For the Ipaddr.t not used, I don't see why it would be a problem.

dinosaure:
Due to #311, this PR will be close. I think we fixed on #311 what we tried to do in this PR. Thanks!

Issue #135, opened 2016-07-18
Conduit need a lot of RAM to build


vbmithr:
1Go is not enough. Even core and async compile fine with this amount of RAM. Why?

avsm:
What options are you building with?  Does this only trigger on the Async backend by any chance?

vbmithr:
On 18/07/2016 23:02, Anil Madhavapeddy wrote:
> What options are you building with? Does this only trigger on the Async
> backend by any chance?

I'm indeed using Async. I'm gonna check if it changes anything on Lwt. I
think this is because the build system uses 2 threads by default, while
the other packages I compiled are only compiled sequentially.

-- 
Vincent

avsm:
yes could be concurrent builds pushing you over.  on ARM?

vbmithr:
On 18/07/2016 23:26, Anil Madhavapeddy wrote:
> yes could be concurrent builds pushing you over. on ARM?

On an EC2 micro instance.

-- 
Vincent

Issue #136, opened 2016-07-18
Conduit is build 2 times


vbmithr:
`make install` causes everything to be rebuilt. So when doing an `opam install conduit`, the compilation is actually performed 2 times.

rgrinberg:
Fixed by #137

PR #137, opened 2016-07-18
fix build

vbmithr:


vbmithr:
This works for me, and fixes #135 and #136

rgrinberg:
Works for me too. We really need something a less sucky build system for conduit though.

PR #138, opened 2016-07-26
add Conduit_lwt_cmdliner for a standard way to retrieve a `mode`

avsm:


opam
```diff
@@ -22,6 +22,9 @@ depends: [
   "uri"
   "cstruct" {>="1.0.1"}
   "ipaddr" {>="2.5.0"}
+  "cmdliner"
+  "logs"
+  "fmt"
```
rgrinberg: Where is `fmt` being used?

avsm: There seems to be a bug in the Fmt_cli meta files (or some other combination) that doesn't pull in the right cmx files and causes a link error if all three aren't specified.  I'll leave this open until I can repro to create an upstream issue.

dbuenzli: Do you have the link error ?

dbuenzli: Also I don't see why you add logs I don't see it being used by this PR.

PR #139, opened 2016-07-27
Add `disconnect` to `FLOW` implementation

djs55:
Related to [mirage/mirage#550]

Signed-off-by: David Scott dave@recoil.org


dinosaure:
This PR is a bit too hold 👍 so I will close it due to #311 and the major release.

Issue #141, opened 2016-07-31
Handle failed lwt threads in server


copy:
Currently failing lwt threads are swallowed (I checked `Conduit_lwt_unix.serve`, but maybe there are other places). I think the error should be reported somehow, or at least this behaviour should be documented. See example below.

```ocaml
let () =
  let ctx = Conduit_lwt_unix.default_ctx in
  let mode = `TCP (`Port 1234) in
  let main = Conduit_lwt_unix.serve ~ctx ~mode (fun flow ic oc -> Lwt.fail_with "oh no") in
  Lwt_main.run main
```

vbmithr:
Hey,

I'm not using the Lwt backend anymore but I agree that failing threads
should never be swallowed. Feel free to push a PR fixing this. Thanks.

avsm:
Right, Lwt really needs some sort of Monitor-style library where errors can be dispatched too. Another option is to use [Logs](http://erratique.ch/software/logs) to record them.

rgrinberg:
Easiest approach would be to follow the async style of providing an error callback to the server. See `on_handler_error`.

copy:
> Easiest approach would be to follow the async style of providing an error callback to the server. See on_handler_error.

I think an easier approach would be to follow the strategy of `Lwt_io.establish_server_safe` and pass all errors to `Lwt.async_exception_hook`, so the caller can wrap the callback into `Lwt.catch` themself.

SGrondin:
Thank you @copy for https://github.com/mirage/ocaml-conduit/pull/143/files . I'm eagerly waiting for it to be merged into master. In the mean time I'm using it in my fork.

PR #143, opened 2016-08-02
 Handle failed lwt threads in server

copy:
This is a fix for #141. It follows the strategy of `Lwt_io.establish_server_safe`, which passes exceptions from the callback to `Lwt.async_exception_hook` (which by default prints the exception and terminates the program).


rgrinberg:
Thank you.

PR #144, opened 2016-08-11
Allow TCP ctx from existing file_descriptors

SGrondin:
This PR opens up the possibility of tracking and manipulating server file descriptors from the application code. I've implemented it for my own needs and I hope others will also find it useful.


SGrondin:
A decision regarding this related PR should probably also be taken. https://github.com/mirage/ocaml-conduit/pull/111/files

SGrondin:
I'm glad to see this being merged in!

Issue #140, opened 2016-07-27
mirage resolver fails with "name resolution failed" when given ip:port based uri


sevenEng:
This may be one of the corner cases, but I tried issuing HTTP requests using ip:port generated uri.

While targeting to unix and using `resolver_unix_system` everything works fine, but switching to xen and [resolver_dns](https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L949) gives

```
Fatal error: exception (Failure "name resolution failed: unknown endpoint type") 
```

I think this exception comes from this function: [dns_stub_resolver](https://github.com/mirage/ocaml-conduit/blob/master/lib/resolver_mirage.ml#L109). Before referring to `DNS.gethostbyname`, if the host part of uri is an ip address, maybe we could just return it and the port. If this is the case, I could submit a patch later.

yomimono:
Fixed by the reporter in #145 - thanks again @sevenEng !

PR #145, opened 2016-08-18
support ip based uri host for mirage resolver

sevenEng:
try to solve issue #140 

it doesn't seem to be an invalid case where we need to issue requests using ip:port (like within local network) rather than using a domain name, however mirage resolver would fail under these circumstances as described in the referenced issue

### DIFF
--- a/lib/resolver_mirage.ml
+++ b/lib/resolver_mirage.ml
@@ -110,8 +110,9 @@ module Make(DNS:Dns_resolver_mirage.S) = struct
       : Conduit.endp Lwt.t =
     let host = get_host uri in
     let port = get_port service uri in
-    DNS.gethostbyname ~server:ns ~dns_port:ns_port dns host
-    >>= fun res ->
+    (match Ipaddr.of_string host with
+    | None -> DNS.gethostbyname ~server:ns ~dns_port:ns_port dns host
+    | Some addr -> return [addr]) >>= fun res ->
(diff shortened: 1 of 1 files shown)


yomimono:
This patch looks good to me.  Any reason not to merge it?

rgrinberg:
@sevenEng This looks good. Any reason why you closed this?

rgrinberg:
@yomimono If you still think this is good, then let's merge it.

yomimono:
Yes, I still think this looks good.  The branch from this PR isn't recoverable ("force-pushed or recreated" according to GitHub) but if @sevenEng would be so good as to resubmit, I'll merge.

samoht:
I think you can still get the commit here: https://github.com/mirage/ocaml-conduit/pull/145/commits/d957c8e138af10da6764d2ff53f053de07b2a53f

rgrinberg:
That's OK, I already cherry picked.

sevenEng:
@rgrinberg Sorry, this was some unexpected side effect of deletion/rebasing/re-push of the branch on my local fork

I'll resubmit again, maybe that's easier for the merge?

rgrinberg:
@sevenEng there's no need, it's in master already https://github.com/mirage/ocaml-conduit/commit/e308d4321b49889328cba52eebeba21425aa4ae3

sevenEng:
Thanks for merging!

PR #151, opened 2016-09-25
Increase backlog for Conduit_lwt_tls and Conduit_lwt_unix_ssl

sepeth:
In #150, I've made changes for Conduit_lwt_unix. @edwintorok suggested to make these changes for Conduit_lwt_tls and Conduit_lwt_unix_ssl, too.

I have one question though. At first, I thought that for consistency I should rename nconn to backlog, but I am not sure how this will affect the consumers? What do you think I should do? Leave it as it is, or rename it?

### DIFF
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -38,11 +38,11 @@ module Client = struct
 end
 
 module Server = struct
-  let listen nconn sa =
+  let listen backlog sa =
     let fd = Lwt_unix.socket (Unix.domain_of_sockaddr sa) Unix.SOCK_STREAM 0 in
     Lwt_unix.(setsockopt fd SO_REUSEADDR true);
     Lwt_unix.bind fd sa;
-    Lwt_unix.listen fd nconn;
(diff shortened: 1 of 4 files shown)


rgrinberg:
My opinion is to rename it to nconn. This is the convention in lwt itself.

sepeth:
Hmm, that's right! The reason I named it as backlog in the first place is that I guess backlog is more commonly used outside, and I think it is a better name for what it represents. It is not the number that represents the number of connections we can have simultaneously, it is the size of the backlog, size of the queue of the items that our application hasn't accepted yet. Also async part uses the name backlog. But it's your call, if it is commonly used parameter and if changing it is going to break others, I can go and fix my previous PR to make this more consistent.

rgrinberg:
Sorry, I misspoke in my previous message. The actual convention is `?backlog` in both Lwt and Async. And we should follow it.

sepeth:
Thanks for quick reply! I will update this quickly ^_^

rgrinberg:
Thank you.

PR #146, opened 2016-08-22
Revamp config step

rgrinberg:
Rewrite it in pure OCaml. Now you just need to run:

```
$ ocaml discover.ml
```

And subsequence `$ make` calls will be incremental.

PR #154, opened 2016-10-10
[followup to #146] refresh build system

avsm:
Testing this before a topkg jump


rgrinberg:
@avsm I'll also cut a release if you don't mind.

PR #156, opened 2016-10-12
SNI (Server Name Indication) for SSL

pdonadeo:
- hostname for SNI (Server Name Indication) added to SSL configuration 
- hostname can be passed to ssl_connect


pdonadeo:
DOH!, it doesn't compile because I submitted a pull request in async_ssl and the tests will fail until that patch will be integrated. Similar situation for cohttp :-(

PR #158, opened 2016-10-24
the API of Tls_mirage.client_of_flow was changed in master

hannesm:
see https://github.com/mirleft/ocaml-tls/pull/336

### DIFF
--- a/lib/conduit_mirage.ml
+++ b/lib/conduit_mirage.ml
@@ -308,7 +308,7 @@ module TLS = struct
 
   let connect (t:t) (`TLS (c, x): client) =
     connect t x >>= fun flow ->
-    TLS.client_of_flow c "??" flow >>= function
+    TLS.client_of_flow c flow >>= function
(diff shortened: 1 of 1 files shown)


hannesm:
(depending on your conduit release plans, this should be hold off till a new tls release is done (likely once MirageOS3 is released))

avsm:
We can cut a release just for this before any other changes I think

hannesm:
I don't have any plans to cut a TLS release too soon (rather wait till happy with the API).

avsm:
Well, we'll have to cut one with the rest of the Mirage3 when the first dev is cut, right?  That's when the conduit release can come out too.

hannesm:
sure.  what I tried to express is that since conduit is used for other things (than MirageOS), there might be a reason to publish a conduit release before Mirage3.  But I don't follow conduit development.

hannesm:
now have a mirage-dev branch, and this in it

PR #161, opened 2016-11-04
try returning an lwt thread that also runs Stack.listen

yomimono:
`tcpip_stack_direct.ml` no longer starts a listening thread automatically with `let _ = S.listen` on connect.  Libraries which wrap the stack need to be sure to call `listen` to allow the stack to hear incoming traffic (even if no TCP or UDP listeners are registered -- replies to outgoing connection requests still need to be heard, and the stack will still need to hear lower-level packets to communicate on the network at all).

### DIFF
--- a/lib/conduit_mirage.ml
+++ b/lib/conduit_mirage.ml
@@ -199,7 +199,8 @@ module TCP (S: V1_LWT.STACKV4) = struct
         let f = Flow.create (module S.TCPV4) flow in
         fn f
       );
-    s
+    let stack_listener = S.listen t in
+    Lwt.pick [stack_listener; s]
 
 end
 


yomimono:
I feel a little out of my depth with this PR, so I'd appreciate a bit of review from someone who's a bit more comfortable with conduit.  @avsm or @rgrinberg  , maybe?

hannesm:
maybe we should have a mirage-dev branch here?  you'll need #158 as well to use conduit with the mirage-dev remote

yomimono:
This PR shouldn't break anything in the current released Mirage ecosystem, FWIW.

yomimono:
There seems to be a test failure on `debian-unstable` related to `ssl`, which should be unaffected by this PR and also appears on #158 .  I think it's unrelated to this patch.

yomimono:
This patch works for conduit servers, but doesn't fix the problem for users which only open outbound connections.  I'll add another commit which addresses this case.

yomimono:
I'm closing this in favor of restoring the previous behavior in `mirage-tcpip`.

PR #163, opened 2016-11-04
Fix shell expansion in Makefile

samoht:
This fix an error introduced in 4ec3f9a9cbc421cca870773775c102c4df748a64
and discovered by @hannesn. I have tested it and seems to work.


hannesm:
actually TravisCI discovered it, I only read logs ;)

PR #164, opened 2016-11-05
Fix shell expansion II / NUL

fdopen:
Follow up to the the latest two commits 😉


samoht:
THANKS!

samoht:
The only failing test is a run of TravisCI failing to install ssl, so unrelated. Merging!

(and thanks again ... I have tried to resurrect my Windows VM without success so was a bit hard to test my change)

PR #167, opened 2016-12-02
fix mirage tests

vbmithr:
Tried to compile mirage tests. Apparently nobody tested that for a little while. I fixed warnings, and so on, but it fails at `make`: seems the `main.ml` is generated invalid and does not compile against latest Mirage. Can anybody help?

PR #169, opened 2016-12-02
followon - get mirage tests building

yomimono:
Builds on #167 .


vbmithr:
Probably I should have pushed in a branch in this repository instead of cloning it…

PR #171, opened 2016-12-05
tests: add an exit test

vbmithr:


rgrinberg:
@vbmithr would you mind addressing my comments in this PR? :/

vbmithr:
On 06/12/2016 15:21, Rudi Grinberg wrote:
> @vbmithr <https://github.com/vbmithr> would you mind addressing my
> comments in this PR? :/

Which comments? I did not see any.

rgrinberg:
Ahh crap, I forgot to hit the submit button. Haven't gotten used to this new interface. Carry on...

Makefile
```diff
@@ -66,5 +66,12 @@ pr:
 	OPAMYES=1 opam publish submit $(NAME).$(VERSION) && rm -rf $(NAME).$(VERSION)
 
 ppx:
+	./configure
```
vbmithr: So that configure is called when I make. Without this, I have to do `./configure` manually when the config does not exit. I put this for convenience given that the configure phase is quick…

vbmithr: But OK, agree that it should only be called when needed, just did not find an easy way to do it.

Makefile
```diff
@@ -3,7 +3,7 @@
 EXT_OBJ:=$(shell ocamlfind ocamlc -config | awk '/^ext_obj:/ {print $$2}')
 EXT_LIB:=$(shell ocamlfind ocamlc -config | awk '/^ext_lib:/ {print $$2}')
 
-OCAMLBUILD = ocamlbuild -use-ocamlfind -classic-display -no-links \
+OCAMLBUILD = ocamlbuild -use-ocamlfind -classic-display \
```
rgrinberg: Also @vbmithr how come you removed `-no-links`? Were they breaking something?

vbmithr:
No I just found them handy. Are they an issue?

rgrinberg:
Nah, it's fine. I see they're useful for tests, but they're annoying for all the binary artifacts they symlink.

PR #172, opened 2016-12-07
Use Logs to report Conduit_lwt_unix_ssl exceptions

talex5:
Before, we silently discarded errors here.

I only added logging to `Conduit_lwt_unix_ssl` for now because I'm trying to debug a problem there at the moment (after a week or so, attempting to connection to my server just hangs, although the process is still working fine apart from that). If people are happy with the extra dependency, I could add logging elsewhere too.

### DIFF
--- a/discover.ml
+++ b/discover.ml
@@ -138,7 +138,7 @@ let vchan       = Flag.mk "vchan" ["vchan"]
 let vchan_lwt   = Flag.mk "vchan_lwt" ["vchan.lwt"]
 let launchd_lwt = Flag.mk "launchd_lwt" ["launchd.lwt"]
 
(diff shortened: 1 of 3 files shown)


vbmithr:
+1. I like small and high quality deps! :)

avsm:
Logging elsewhere would be appreciated!

rgrinberg:
@talex5 Note that this is also catching and logging errors from the server callback itself and not just the accept call. Is that intended?

talex5:
@rgrinberg It is, but perhaps the wording could be improved. In the cohttp case (and in most cases probably) `process_accept` will run the user's callback in a new thread, with its own error handling. But maybe this should say the error was while "handling a client connection" rather than while "accepting" it.

PR #173, opened 2016-12-08
Copy server stop fixes + Steps towards conduit server notifications

rgrinberg:
This started off as a port of server stop fixes we just merged and ended up as a unification of a lot of conduit/lwt server code. This code still isn't good, but at least we won't have to fix every bug 3 times and iterate on it in 1 place.

### DIFF
--- a/lib/conduit_lwt_tls.mli
+++ b/lib/conduit_lwt_tls.mli
@@ -32,8 +32,10 @@ module Server : sig
     Tls.Config.server ->
     Lwt_unix.file_descr ->
     (Lwt_unix.file_descr * Lwt_io.input_channel * Lwt_io.output_channel) Lwt.t
+  [@@@deprecated "Low level function. Will be removed."]
 
   val listen : int -> Lwt_unix.sockaddr -> Lwt_unix.file_descr
+  [@@@deprecated "Low level function. Will be removed."]
 
   val init :
     ?backlog:int ->
--- a/lib/conduit_lwt_tls.mli
+++ b/lib/conduit_lwt_tls.mli
@@ -46,4 +46,16 @@ module Server : sig
     Lwt_unix.sockaddr ->
     (Lwt_unix.file_descr -> Lwt_io.input_channel -> Lwt_io.output_channel -> unit Lwt.t) ->
     unit Lwt.t
+
+  val init'
+    : ?backlog:int
+    -> ?stop:(unit Lwt.t)
+    -> ?timeout:int
+    -> Tls.Config.server
+    -> Lwt_unix.sockaddr
+    -> (Lwt_unix.file_descr
+        -> Lwt_io.input_channel
+        -> Lwt_io.output_channel
+        -> unit Lwt.t)
+    -> unit Lwt.t
 end
--- a/lib/conduit_lwt_tls.mli
+++ b/lib/conduit_lwt_tls.mli
@@ -33,8 +33,7 @@ module Server : sig
     Lwt_unix.file_descr ->
     (Lwt_unix.file_descr * Lwt_io.input_channel * Lwt_io.output_channel) Lwt.t
 
-  val listen :
-    int -> Lwt_unix.sockaddr -> Lwt_unix.file_descr
+  val listen : int -> Lwt_unix.sockaddr -> Lwt_unix.file_descr
 
   val init :
     ?backlog:int ->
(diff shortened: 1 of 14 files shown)


rgrinberg:
That last commit. I have no idea how that Lwt.task made it into the codebase since the original diff in #111  showed that line as removed. @vbmithr did you merge things incorrectly perhaps?

rgrinberg:
@avsm any idea on what's wrong with appveyor?

rgrinberg:
@yomimono @avsm @samoht Do you agree with the deprecations? None of those functions seem particularly useful for end users. But I'd like to make sure they aren't being used downstream somewhere in  mirage.

rgrinberg:
OK the scope for this PR is already quite big. I have a couple of other good changes designed to address all the duplication but let's make sure this all works first. I did try to change the functionality as little as possible.

@vbmithr @edwintorok Hopefully this PR is a better base to rebase #116 on top of. I will improve this further as well.

@talex5 your logging is now available for every lwt server. It would be best if the logger knew which type of server it's running though. We'll address that in the future hopefully.

lib/conduit_lwt_unix.ml
```diff
@@ -175,56 +175,33 @@ module Sockaddr_client = struct
...
-        return (fd, ic, oc)
+        Lwt.return (fd, ic, oc)
       )
 end
 
 module Sockaddr_server = struct
-  open Conduit_lwt_server
-
-  let init_socket ?(backlog=128) sockaddr =
-    Unix.handle_unix_error (fun () ->
```
rgrinberg: @avsm I think you added this line originally. I had to get rid of it since the other Lwt backends didn't use it. I don't think exiting the program is good behavior anyway.

lib/conduit_lwt_tls.mli
```diff
@@ -32,9 +32,10 @@ module Server : sig
     Tls.Config.server ->
     Lwt_unix.file_descr ->
     (Lwt_unix.file_descr * Lwt_io.input_channel * Lwt_io.output_channel) Lwt.t
+  [@@@deprecated "Low level function. Will be removed."]
```
talex5: Is this function is the only way to pass in a `Tls.Config.server`? That seems quite important.

vbmithr: If they are removed `init` must provide a way to pass a `server`. I think the functions in this module overall should leave more choice to the user (i.e. do not force "no authentication" to the client…). Since users can directly access this module it does not have to have the same interface as `conduit_lwt_unix`.

vbmithr:
@avsm @samoht Do we need to wait for Appveyor to be fixed or can we merge this?

lib/conduit_lwt_tls.mli
avsm: Perhaps point to the function that should be used, in the deprecation warning?

lib/conduit_lwt_unix.ml
avsm: Agreed!

tests/mirage/vchan/unikernel.ml
```diff
@@ -24,7 +24,7 @@ module Server(Time : V1_LWT.TIME) = struct
         | `Eof ->
           Log.info (fun f -> f "End of transmission!"); Lwt.return_unit
         | `Error e ->
-          Log.warn (fun f -> f "Error reading the vchan flow!");
+          Log.wlarn (fun f -> f "Error reading the vchan flow!");
```
avsm: typo?

avsm:
The Appveyor failure is due to a recent regression in Cygwin with a new diffutils version.  @samoht, you worked around this elsewhere didnt you?

samoht:
The fix for appveyor si to remove `-P diffutils` from the init line.

samoht:
@rgrinberg in general I am ok with deprecating functions. Would be useful to tell the user what they should use instead, though. I guess we need better `init` functions in all cases, as @vbmithr suggested.

lib/conduit_lwt_tls.mli
rgrinberg: So there seems to be no way to provide a `Tls.Config.server` to a new conduit server. Unless you're writing a TCP server from scratch and use this accept function in your accept loop.

How about adding a new version of init that does provide this functionality? The current init version can be implemented on top of that and eventually removed.

@avsm Will do.

rgrinberg:
@samoht @vbmithr @talex5 The latest commit contains a version of init that accepts a Tls server config. I don't think of the other backends had such restrictions.

rgrinberg:
Will merge this tomorrow unless anyone objects.

PR #175, opened 2016-12-10
Remove dependency to stringext

vbmithr:


vbmithr:
@grinberg: Please review :)

rgrinberg:
But why remove it? I don't have a problem if we're switching to astring or the functionality is covered elsewhere. But just copy pasting the same couple of functions across opam package I'm against.

So I'm -1 on this, unless someone has a concrete problem with stringext.

vbmithr:
On 10/12/2016 17:45, Rudi Grinberg wrote:
> So I'm -1 on this, unless someone has a concrete problem with stringext.

Just to remove a dep since we use it on only one occurrence in this
code. If we happen to need more String utilities, then let's add astring
in the future?


-- 
Vincent

rgrinberg:
My position on this issue doesn't depend on the number of functions that we'd need to use. It's a matter of replacing a tested implementation with an untested one for benefits that are entirely unclear to me.

But regardless, I'm not the only maintainer of conduit so I'll defer this to @avsm @samoht if they'd like to go ahead with this.

vbmithr:
On 10/12/2016 20:02, Rudi Grinberg wrote:
> My position on this issue doesn't depend on the number of functions that
> we'd need to use. It's a matter of replacing a tested implementation
> with an untested one for benefits that are entirely unclear to me.
> 
> But regardless, I'm not the only maintainer of conduit so I'll defer
> this to @avsm <https://github.com/avsm> @samoht
> <https://github.com/samoht> if they'd like to go ahead with this.
> 

From my point of view, removing a dep when a project uses only one
simple function on one instance is a good thing, hence my patch.

Ok to replace with astring if it assumed that it will be useful in the
future.

-- 
Vincent

avsm:
I would prefer a proper switch over to Astring, since that's the string handling interface of choice now in Mirage (in terms of general use, not just here in ocaml-conduit).

vbmithr:
On 30/12/2016 16:09, Anil Madhavapeddy wrote:
> I would prefer a proper switch over to Astring, since that's the string handling
> interface of choice now in Mirage (in terms of general use, not just here in
> ocaml-conduit).

Fine, agreed.

avsm:
(and with many thanks to @rgrinberg for creating the very useful Stringext in the first place -- the situation before was a total mess). We're getting there :-)

PR #177, opened 2016-12-12
Fix native TLS

vbmithr:
@hannesm This looks an improvement. I now have this: `Fatal error: exception Tls_lwt.Tls_alert(14)`. It looks it is a certificate error, however my certificate looks right…

### DIFF
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -56,7 +56,7 @@ module Server = struct
           (fun () -> Tls_lwt.Unix.server_of_fd tls fd)
           (fun t ->
              let (ic, oc) = Tls_lwt.of_t t in
-             Lwt.return (fd, ic, oc))
+             callback fd ic oc)
           (fun exn -> Lwt_unix.close fd >>= fun () -> Lwt.fail exn)
         |> Lwt.ignore_result)
 


rgrinberg:
This is an obvious bug, thanks. I wonder how we can get decent reporting back meanwhile.

vbmithr:
On 12/12/2016 20:42, Rudi Grinberg wrote:
> This is an obvious bug, thanks. I wonder how we can get decent reporting
> back meanwhile.

We should setup CI testing for servers, probably.

-- 
Vincent

rgrinberg:
Err sorry, I actually meant wranings. It seems like the way we use ppx_optcomp doesn't play well with unused variables for example. I wonder if switching to cppo meanwhile would help.

vbmithr:
Yeah. Let's face it. We cannot develop under in such conditions.

hannesm:
this looks like a good bug fix... sorry I don't understand the alert 14, neither what 14 tries to encode, nor what "the certificate looks right" should mean.

Our TLS implementation does some pedantic checks of server certificates: at least 1024 bit RSA, there must be a chain of trust between the top certificate in the given chain and the server certificate, it must have the KeyUsage extension (set to sign/encipherment depending on which key exchange you use); if supplying multiple certificates (to use SNI), the set of names may not overlap.  I don't know the origin of your certificate, neither the type nor intended usage.

I agree that `Tls_lwt.Tls_alert(14)` should be improved.  Will try to find some time for this.

vbmithr:
Ok that was a lame assertion "the certificate looks right". I'm gonna try to produce a certificate that match your requirements. Thanks :)

vbmithr:
Ok, so this works, it was just a "normal" exn raised when the client cut the connection (here, firefox).

PR #181, opened 2016-12-15
serve: change error mgmt

vbmithr:
For Tezos, we need a way that server exception are not sent to `async_exception_hook`, because we want to filter them otherwise we could not differentiate them from similar exceptions coming from another part of the program.

### DIFF
--- a/lib/conduit_lwt_unix.mli
+++ b/lib/conduit_lwt_unix.mli
@@ -156,18 +156,19 @@ val init : ?src:string -> ?tls_server_key:tls_server_key -> unit -> ctx io
     via the [ctx] context to the endpoint described by [client] *)
 val connect : ctx:ctx -> client -> (flow * ic * oc) io
 
-(** [serve ?backlog ?timeout ?stop ~ctx ~mode fn] establishes a
-    listening connection of type [mode], using the [ctx] context.
-    The [stop] thread will terminate the server if it ever becomes
-    determined.  Every connection will be served in a new
-    lightweight thread that is invoked via the [fn] callback.
-    The [fn] callback is passed the {!flow} representing the
-    client connection and the associated input {!ic} and output
-    {!oc} channels. If the callback raises an exception, it is
-    passed to [!Lwt.async_exception_hook]. *)
+(** [serve ?backlog ?timeout ?stop ?on_exn ~ctx ~mode fn]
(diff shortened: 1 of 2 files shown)


rgrinberg:
Looks fine, but I think this hook should return a `unit`. This should signify to users that we are never waiting for its result.

Also I'd prefer a name like `on_exn`, but perhaps I'm bikeshedding that.

rgrinberg:
This looks good to me. I'll wait for a couple of days to give others a chance to review and unless there's feedback will merge.

lib/conduit_lwt_unix.mli
```diff
@@ -156,18 +156,19 @@ val init : ?src:string -> ?tls_server_key:tls_server_key -> unit -> ctx io
...
-(** [serve ?backlog ?timeout ?stop ~ctx ~mode fn] establishes a
-    listening connection of type [mode], using the [ctx] context.
-    The [stop] thread will terminate the server if it ever becomes
-    determined.  Every connection will be served in a new
-    lightweight thread that is invoked via the [fn] callback.
-    The [fn] callback is passed the {!flow} representing the
-    client connection and the associated input {!ic} and output
-    {!oc} channels. If the callback raises an exception, it is
-    passed to [!Lwt.async_exception_hook]. *)
+(** [serve ?backlog ?timeout ?stop ?exn_hook ~ctx ~mode fn]
```
yomimono: If `exn_hook` is renamed in `conduit_lwt_unix.ml`, it should be here too.

lib/conduit_lwt_unix.mli
```diff
@@ -156,18 +156,19 @@ val init : ?src:string -> ?tls_server_key:tls_server_key -> unit -> ctx io
...
+    establishes a listening connection of type [mode], using the [ctx]
+    context.  The [stop] thread will terminate the server if it ever
+    becomes determined.  Every connection will be served in a new
+    lightweight thread that is invoked via the [fn] callback.  The
+    [fn] callback is passed the {!flow} representing the client
+    connection and the associated input {!ic} and output {!oc}
+    channels. If the callback raises an exception, it is passed to
+    [exn_hook] (by default, to !Lwt.async_exception_hook). *)
```
yomimono: Another dangling `exn_hook` that should be `on_exn`.

vbmithr:
Thanks Mindy :)

PR #182, opened 2016-12-22
Adapt to the new error scheme + depends on mirage-flow instead of mirage-types

samoht:


samoht:
Seems that Travis is not testing the mirage backend, but merging anyway as it seems to be working when used by revdeps.

PR #186, opened 2016-12-28
Tests: do not link with lwt.ssl if it is not installed

edwintorok:
cdtest.ml uses lwt.ssl, and the Makefile won't build it if lwt.ssl is
not available.
cdtest_tls works fine with just tls.lwt, but if conduit got built with
lwt.ssl, then lwt.ssl is needed during linking, otherwise it is not.

Fixes https://github.com/ocaml/opam-repository/issues/8166

All the travis tests have `ssl` as depopts installed, I wonder if its possible to have at least one Travis build which has only `tls.lwt'` installed but not `ssl`.


rgrinberg:
Thanks. Will tweak this a little to omit the empty line. And then make a point release.

PR #191, opened 2016-12-30
Fix discover.ml

edwintorok:
/cc @rgrinberg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ env:
  global:
    - PACKAGE="conduit"
    - DEPOPTS="async_ssl lwt mirage-dns ssl tls"
-   - FORK_BRANCH=HEAD
  matrix:
    - DISTRO=debian-stable OCAML_VERSION=4.02.3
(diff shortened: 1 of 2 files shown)


rgrinberg:
Haha, that is very embarrassing.

edwintorok:
I'm thinking about adding a few more tests to the matrix (in a separate PR), obviously we can't test all 2^depopts combinations, but having these 2 combinations should cover the main ones: DEPOPTS=`lwt ssl` DEPOPTS=`lwt tls`. What do you think?

rgrinberg:
I agree on expanding it but your 2 main ones exclude mirage and async so it's not really that extensive.

@avsm what do you think?

edwintorok:
Should we also have: `DEPOPTS="async"` and `DEPOPTS="lwt mirage-dns"`? (in addition to the current one, which I'd of course keep: `DEPOPTS="async_ssl lwt mirage-dns ssl tls"`)

rgrinberg:
I think that would be pretty complete then @edwintorok

PR #194, opened 2017-01-02
Use random ports in unix tests

rgrinberg:
cc @edwintorok

### DIFF
--- a/tests/unix/cdtest.ml
+++ b/tests/unix/cdtest.ml
@@ -16,7 +16,10 @@
 
 open Lwt.Infix
 
-let port = 9192
+let port =
+  Random.self_init ();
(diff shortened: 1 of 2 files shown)


tests/unix/cdtest.ml
```diff
@@ -16,7 +16,7 @@
 
 open Lwt.Infix
 
-let port = 9192
+let port = 16_384 + Random.int 10_000
```
edwintorok: Was about to say this looks good, but then this always returns 25728. I think you need to add this before it: `let () = Random.self_init ()`

edwintorok:
Thanks, if you add `let () = Random.self_init ()` before it then this looks good.

Issue #187, opened 2016-12-29
opam install conduit -t fails


dbuenzli:
```
opam install conduit -t
The following actions will be performed:
  ∗  install conduit 0.14.0
  ∗  install cohttp  0.21.0
  ∗  install github  2.1.0 
===== ∗  3 =====
Do you want to continue ? [Y/n] y

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫 
[cohttp] Archive in cache
[conduit] Archive in cache
[github] Archive in cache

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫 
[ERROR] The compilation of conduit failed at "make test".
Processing  1/3: [conduit: ocamlfind remove]
#=== ERROR while installing conduit.0.14.0 ====================================#
# opam-version 1.2.2
# os           darwin
# command      make test
# path         /Users/dbuenzli/.opam/4.03.0/build/conduit.0.14.0
# compiler     4.03.0
# exit-code    2
# env-file     /Users/dbuenzli/.opam/4.03.0/build/conduit.0.14.0/conduit-99191-d2d111.env
# stdout-file  /Users/dbuenzli/.opam/4.03.0/build/conduit.0.14.0/conduit-99191-d2d111.out
# stderr-file  /Users/dbuenzli/.opam/4.03.0/build/conduit.0.14.0/conduit-99191-d2d111.err
### stdout ###
# [...]
# ocamlfind ocamldep -package ipaddr.unix -package lwt.ssl -package lwt.unix -package tls.lwt -package sexplib -package ipaddr -package cstruct -package uri -package stringext -package logs -modules tests/unix/cdtest.ml > tests/unix/cdtest.ml.depends
# ocamlfind ocamlc -c -w A-4-33-40-41-42-43-34-44 -g -bin-annot -short-paths -principal -thread -strict-sequence -package ipaddr.unix -package lwt.ssl -package lwt.unix -package tls.lwt -package sexplib -package ipaddr -package cstruct -package uri -package stringext -package logs -I tests/unix -I lib -I tests/mirage -I tests/async -o tests/unix/cdtest.cmo tests/unix/cdtest.ml
[...]

PR #200, opened 2017-01-03
Prepare 0.14.4

rgrinberg:
two releases in 1 day.

Very agile!


samoht:
Does it fix the test breakage as well? If not, maybe we should just disable the tests in the release versions?

rgrinberg:
It fixes building the tests. I think some people may still have issues running them #187 but those seem to be more rare.

edwintorok:
I wouldn't mind if tests are disabled in opam until we can actually run them from Travis.

samoht:
See https://github.com/ocaml/opam-repository/pull/8197

rgrinberg:
OK, thanks for fixing opam-repository.

@edwintorok what's preventing us from running the tests in travis?

edwintorok:
I haven't had time to work on this yet: https://github.com/ocaml/ocaml-ci-scripts/pull/145. Basically we need a way to tell opam to run the tests for just 1 package, not for every of it dependencies too. One way to do that is to install all dependencies without `-t`, and then do a `-t` on the package. However this would also need to parse a dry-run/json output of opam to figure out what those packages are (using `jq` (or something else with minimal dependencies)). I tried a simpler solution that just did `opam install --deps-only package` install followed by a `opam install -t package`, but that still installs and runs the tests of some test-only dependencies. This in turn breaks the CI of other projects (because their test-only deps' tests may not work).

talex5:
https://github.com/ocaml/opam/issues/2277#issuecomment-250056045 and https://github.com/ocaml/opam/issues/2190 seem to suggest that the `-t` problem is fixed in opam 2.

rgrinberg:
OK I've disabled the tests as recommended.

PR #206, opened 2017-01-21
Fix lwt server exception handling

### LINKED ISSUE #204: Conduit silently ignores exceptions from handlers
```
#require "conduit.lwt-unix";;

open Lwt.Infix

let () =
  Logs.set_reporter (Logs_fmt.reporter ())

let addr = Unix.ADDR_INET (Unix.inet_addr_loopback, 8080)

let callback _ =
  print_endline "In callback";
  Lwt.fail (Failure "An error!")

let () =
  Lwt_main.run begin
    Conduit_lwt_server.listen addr
    |> Conduit_lwt_server.init callback
  end
```

If you run this program and try to connect to `localhost:8080`, it prints `In callback` but the exception is not displayed.

I think the problem is here, in `Conduit_lwt_server`:

```
         | `Accept v ->
           Lwt.on_termination (handler v) disconnected;
           loop ()
```

`on_termination` will call `disconnected` if the handler fails, but will not report the error anywhere.

Also, if `callback` raises an exception without blocking (e.g. using `failwith` instead of `Lwt.fail`) then the server process exits prints the exception and then exits because the `Log.warn` doesn't call `loop`.

rgrinberg:
Don't terminate conduit on exceptions. Log handler exceptions

Fix #204

cc @edwintorok

### DIFF
--- a/lib/conduit_lwt_server.ml
+++ b/lib/conduit_lwt_server.ml
@@ -71,14 +71,15 @@ let init ?(stop = fst (Lwt.wait ())) handler fd =
       (function
         | `Stop -> disconnected (); Lwt.return_unit
         | `Accept v -> loop v)
-      (function
-        | Lwt.Canceled -> disconnected (); Lwt.return_unit
-        | ex ->
-          disconnected ();
-          Log.warn (fun f ->
-              f "Uncaught exception accepting connection: %s"
-                (Printexc.to_string ex));
-          Lwt_unix.yield () >>= accept)
+      (fun exn ->
+         disconnected ();
+         match exn with
+         | Lwt.Canceled -> Lwt.return_unit
+         | ex ->
+           Log.warn (fun f ->
+               f "Uncaught exception accepting connection: %s"
+                 (Printexc.to_string ex));
+           Lwt_unix.yield () >>= accept)
   and loop v =
     Lwt.try_bind
       (fun () -> handler v)
--- a/lib/conduit_lwt_server.ml
+++ b/lib/conduit_lwt_server.ml
@@ -56,26 +56,34 @@ let set_max_active max_active =
   Lwt_condition.broadcast cond ()
 
 let init ?(stop = fst (Lwt.wait ())) handler fd =
-  let stop = Lwt.map (fun () -> `Stop) stop in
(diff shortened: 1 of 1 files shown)


vbmithr:
It should be possible to provide an `~on_exn` callback where exceptions would be passed. I think exceptions should not be swallowed, regardless or whether or not we do it silently (i.e. printing an error msg or not).

rgrinberg:
I agree that such on_exn makes sense. But how do we separate accept vs. callback exceptions in such a callback?

I think we're only "swallowing" accept exceptions here however. In the sense that our users cannot handle them their own way. But I don't see how we can avoid swallowing exception as default behavior however, I think it's very undesirable for our servers to stop working on raises.

lib/conduit_lwt_server.ml
```diff
@@ -56,26 +56,34 @@ let set_max_active max_active =
...
-                (Printexc.to_string ex)
-            );
-          Lwt_unix.yield ()) in
-  Lwt.finalize loop (fun () -> Lwt_unix.close fd)
+                (Printexc.to_string ex));
+          Lwt_unix.yield () >>= accept)
+  and loop v =
+    Lwt.catch
+      (fun () ->
+         Lwt.on_termination (handler v) disconnected;
```
talex5: This seems wrong. `on_termination` will ignore any exception from `handler v`. That needs to go somewhere.

lib/conduit_lwt_server.ml
```diff
@@ -56,26 +56,34 @@ let set_max_active max_active =
...
+         accept ())
```
talex5: This isn't a tail-call here. Every time around the loop you'll add an extra exception handler.

talex5: (`Lwt.try_bind` would solve the immediate problem, but I suspect we only have this problem because the bit above is wrong)

rgrinberg:
@talex5 I think I fixed all the tail recursion problems. Also fixed some missing disconnected calls.

talex5:
LGTM

rgrinberg:
Good point about the switch argument by the way. Should we ask @antron?

PR #207, opened 2017-01-24
not yet mirageos 3 compatible

yomimono:


yomimono:
CentOS and Fedora 24 appear to be having some difficulties talking to the external solver, so I'm going to call those test failures false negatives and merge this.

PR #210, opened 2017-02-03
fix memory leak

vbmithr:
cc @pdonadeo cc @rgrinberg

### DIFF
--- a/lib/conduit_async_ssl.ml
+++ b/lib/conduit_async_ssl.ml
@@ -20,10 +20,9 @@ open Core.Std
 open Async.Std
 open Async_ssl.Std
 
-let ssl_connect ?version ?name ?ca_file ?ca_path ?session ?verify
-    net_to_ssl ssl_to_net =
-  let net_to_ssl = Reader.pipe net_to_ssl in
-  let ssl_to_net = Writer.pipe ssl_to_net in
+let ssl_connect ?version ?name ?ca_file ?ca_path ?session ?verify r w =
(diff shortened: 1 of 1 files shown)


vbmithr:
This seems to at least mitigate the memory leak issue I was experiencing using `cohttp/async`.
I'm pretty sure there was an issue when using `async_ssl` due to closing an `ssl_connect` pipe not cleaning up the underlying writer. I would be happy if anybody can review this.

lib/conduit_async_ssl.ml
```diff
@@ -45,15 +44,27 @@ let ssl_connect ?version ?name ?ca_file ?ca_path ?session ?verify
   >>= fun conn ->
   verify_connection conn >>= function
   | false ->
+    Ssl.Connection.close conn ;
+    Pipe.close_read app_rd ;
+    Writer.close w >>= fun () ->
```
rgrinberg: How come you're closing the reader in 1 case and the pipe in the other? Seems a bit asymmetrical.

rgrinberg:
This is a good change that I've experimented with myself, but I don't think it will affect the memory leak one way or the other. This seems to be a fix for a potential FD leak. But I don't anyone observed FD leaks with this library yet.

vbmithr:
On 03/02/2017 15:31, Rudi Grinberg wrote:
> How come you're closing the reader in 1 case and the pipe in the other?
> Seems a bit asymmetrical.

The other is automatically closed by Ssl.Connection.close closing its
pipes. See related Async code.

PR #211, opened 2017-02-05
Async with connection

vbmithr:


vbmithr:
@rgrinberg : ping?
Could we merge this? I think it was important (especially 873dda0)

PR #212, opened 2017-03-04
mirage-dev

rgrinberg:
@yomimono


yomimono:
@rgrinberg the `build` section for `mirage-conduit.opam` is a bit suspicious:

```
build:   ["ocamlfind" "query" "conduit.mirage"]
```

Let's improve this state of affairs in a separate PR.  I've reverted the `opam` changes and replaced them with a commit that just updates the versions in `mirage-conduit.opam`'s dependencies.

rgrinberg:
OK. But that build step is exactly what we expect. This is the hacky approach to multi package splits we usually take if we our META file is shared between the opam packages. See tyxml for example

PR #213, opened 2017-03-10
Add a way to start a server from an already open fd

samoht:


SGrondin:
How does this relate to https://github.com/mirage/ocaml-conduit/pull/144 ?

samoht:
That seems pretty similar indeed! I am not very happy with my PR, as it is not clear in which state the file descriptor that you pass to the server is. I ended up re-implementing the server logic for my use-case, instead of relying on conduit. Your approach is probably better.

NightBlues:
Its currently not possible to reuse existing socket, so things like STARTTLS doesn't work :(
For example: https://tools.ietf.org/html/rfc3207 https://tools.ietf.org/html/rfc2595

dinosaure:
@NightBlues what is your current case? In other side, we want to implement `STARTTLS` in another project (experimental) here: https://github.com/dinosaure/colombe. So I'm curious about your task 👍 !

NightBlues:
@dinosaure I just need email notifications for my web site :) Unfortunately I was not able to find your library. I've found ocamlnet implementation that seems to be not compatible with lwt. So, we decided to use conduit-lwt with https://github.com/vbmithr/ocaml-smtp

dinosaure:
Ok no worries, `colombe` still is very experimental but I'm currently focus on a client implementation which should follow RFC 6409 at the beginning. I will ping you when I have stable version :) !

PR #214, opened 2017-04-19
Update to lwt.3.0.0

djs55:
Lwt_unix.bind now returns a promise.

Note the signature of `Conduit_lwt_server.listen` has had to change, to also return a promise.

Rleated to ocsigen/lwt#308

Signed-off-by: David Scott <dave@recoil.org>

### DIFF
--- a/lib/conduit_lwt_server.ml
+++ b/lib/conduit_lwt_server.ml
@@ -15,7 +15,7 @@ let close (ic, oc) =
 let listen ?(backlog=128) sa =
   let fd = Lwt_unix.socket (Unix.domain_of_sockaddr sa) Unix.SOCK_STREAM 0 in
   Lwt_unix.(setsockopt fd SO_REUSEADDR true);
-  Lwt_unix.bind fd sa;
+  Lwt_unix.Versioned.bind_1 fd sa;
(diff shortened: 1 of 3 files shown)


rgrinberg:
Doesn't Lwt have a versioned module that will allow us to stay compatible with Lwt <3.0?

aantron:
Indeed, it does, though you will get annoying deprecation warnings on Lwt 2.7.0+, which you can individually suppress, or just bear them until 3.0.0+ is used by (almost) everyone.

You want `Lwt_unix.Versioned.bind_1` for that purpose.

aantron:
I recommend leaving them, though, as a reminder to self. If you switch to the new API and require Lwt 3.0.0 months from now, it will allow Lwt to drop the deprecated `bind_1` at some even farther future date, which may or may not ever come, without suddenly breaking `conduit`.

djs55:
I've updated the patch to use the `bind_1` version (i.e. the original)

rgrinberg:
Hmm, CI is failing:

```
# ocamlfind: Package `dns.lwt-core' not found - required by `dns.mirage'
```

@avsm @samoht Anyways, it looks unrelated so I will merge.

PR #219, opened 2017-05-02
Lwt 3.0 compat of TLS/SSL

rgrinberg:
0.15.3 will be released with this shortly @hcarty

Fixes #219

### DIFF
--- a/lib/conduit_lwt_tls.ml
+++ b/lib/conduit_lwt_tls.ml
@@ -25,7 +25,7 @@ module Client = struct
         let () =
           match src with
           | None -> ()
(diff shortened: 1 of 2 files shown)


hcarty:
Thank you!

hcarty:
Is there a possibility of a race condition here that moving to `Lwt_unix.Versioned.bind_2` could help with?  If I get time I'll test it.

rgrinberg:
I'm not sure about a race condition but it's sensible change anyway. The
only problem is that it's breaking. It will come in 0.16.0

PR #222, opened 2017-05-04
Correct the Findlib package name of lwt_ssl

dkim:
OPAM installs lwt_ssl under the Findlib package name of `lwt_ssl`, not `lwt.ssl`. Refer to https://github.com/ocaml/opam-repository/blob/master/packages/lwt_ssl/lwt_ssl.1.0.1/opam#L24
```
remove: [
    ["ocamlfind" "remove" "lwt_ssl"]
]
```

 and https://ocsigen.org/lwt/api/Lwt_ssl:

> This module is provided by OPAM package `lwt_ssl`. Link with ocamlfind package `lwt_ssl`.

The use of the wrong name prevented `discover.ml` from detecting the installed lwt_ssl package.

### DIFF
--- a/discover.ml
+++ b/discover.ml
@@ -86,7 +86,7 @@ end
 let async       = Flag.mk "async" ["async"]
 let async_ssl   = Flag.mk "async_ssl" ["async_ssl"]
 let lwt         = Flag.mk "lwt" ["lwt"]
-let lwt_ssl     = Flag.mk "lwt_ssl" ["lwt.ssl"]
+let lwt_ssl     = Flag.mk "lwt_ssl" ["lwt_ssl"]
 let lwt_tls     = Flag.mk "lwt_tls" ["tls.lwt"]
 let mirage      = Flag.mk "mirage" ["mirage-types-lwt"; "mirage-flow-lwt"; "dns.mirage"]
 let mirage_tls  = Flag.mk "mirage_tls" ["tls"; "tls.mirage"]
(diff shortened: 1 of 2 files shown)


rgrinberg:
@aantron this will make conduit lwt 3.0 only right?

hcarty:
@rgrinberg Should be on 2.7.0 or later - see http://opam.ocaml.org/packages/lwt_ssl/lwt_ssl.1.0.0/

hcarty:
* but it will also require a dependency on the lwt_ssl opam package

aantron:
@hcarty is right (thank you). `lwt_ssl` was added in 2.7.0 as both an OPAM and Findlib package, and you will need an `opam` dependency on it.

Details are:
- Since 2.7.0, `lwt_ssl` is just an alias for `lwt.ssl`. This is for compatibility during transition away from `lwt.ssl`.
- Since 3.0.0, `lwt.ssl` does not exist anymore, and its contents are now properly in `lwt_ssl`.

rgrinberg:
@dkim please add an opam dependency on lwt_ssl

dkim:
@rgrinberg I have added `lwt_ssl` into the `depopts` field.

mattjbray:
Any chance of a `v0.15.4` release with this fix? As far as I can tell, it's not possible to get `Conduit_lwt_unix_ssl` and Lwt 3.0.0 with the current `v0.15.3` on Opam.

PR #223, opened 2017-05-19
Also pass context certificate to clients.

paurkedal:
This is a sketch of how we might add TLS client authentication. We may want to revise it before merging:

In the server, the local certificate is passed though the `server` type though `server_tls_config`. If I were to do the same with the `client_tls_config` type, I would break backwards compatibility due to the concrete type. The alternative of adding more constructors to `client` seems inelegant. The current patch avoids compatibility issues but only allows passing the key though the context.

Also note, I renamed the type and keyword argument `tls_server_key` to `tls_own_key`. I considered "peer" but "own" makes it clear that it's the local peer and the word was used in `Tls.Config`. But more essentially, we could have separated the server and client keys in the context. The type might still be the same. I left aliases `tls_server_key` for backwards compatibility. (If we wanted to select between multiple keys, we could consider passing a function which receives information about the connection which is about to be established.)

### DIFF
--- a/lwt-unix/conduit_lwt_tls.ml
+++ b/lwt-unix/conduit_lwt_tls.ml
@@ -24,7 +24,7 @@ module Client = struct
     Conduit_lwt_server.with_socket sa (fun fd ->
         (match src with
          | None -> Lwt.return_unit
-         | Some src_sa -> Lwt_unix.Versioned.bind_2 fd src_sa) >>= fun () ->
+         | Some src_sa -> Lwt_unix.bind fd src_sa) >>= fun () ->
         X509_lwt.authenticator `No_authentication_I'M_STUPID >>= fun authenticator ->
         let config = Tls.Config.client ~authenticator ?certificates () in
         Lwt_unix.connect fd sa >>= fun () ->
--- a/lwt-unix/conduit_lwt_tls.ml
+++ b/lwt-unix/conduit_lwt_tls.ml
@@ -20,13 +20,13 @@ open Lwt.Infix
 let _ = Nocrypto_entropy_lwt.initialize ()
 
 module Client = struct
-  let connect ?src host sa =
+  let connect ?src ?certificates host sa =
     Conduit_lwt_server.with_socket sa (fun fd ->
         (match src with
          | None -> Lwt.return_unit
-         | Some src_sa -> Lwt_unix.bind fd src_sa) >>= fun () ->
+         | Some src_sa -> Lwt_unix.Versioned.bind_2 fd src_sa) >>= fun () ->
         X509_lwt.authenticator `No_authentication_I'M_STUPID >>= fun authenticator ->
-        let config = Tls.Config.client ~authenticator () in
+        let config = Tls.Config.client ~authenticator ?certificates () in
         Lwt_unix.connect fd sa >>= fun () ->
         Tls_lwt.Unix.client_of_fd config ~host fd >|= fun t ->
         let ic, oc = Tls_lwt.of_t t in
--- a/lwt-unix/conduit_lwt_tls.mli
+++ b/lwt-unix/conduit_lwt_tls.mli
@@ -21,6 +21,7 @@ module Client : sig
 
   val connect :
     ?src:Lwt_unix.sockaddr ->
+    ?certificates:Tls.Config.own_cert ->
     string ->
     Lwt_unix.sockaddr ->
     (Lwt_unix.file_descr * Lwt_io.input_channel * Lwt_io.output_channel) Lwt.t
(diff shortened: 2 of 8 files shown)


paurkedal:
Forgot, this will fix #7.

paurkedal:
Did a merge with current master, which succeeded with some three-way merges.  It still works on my end (with some unrelated tweaks in conduit and cohttp due to evolving libraries).

paurkedal:
I should have mentioned that I only tested with OpenSSL yesterday. The native fails with `Tls_lwt.Tls_alert(6)` there is no error code 6 in `Tls.Packet.alert_type`). But this is probably an issue for ocaml-tls, it's probaly a lack of overlapping ciphers.

paurkedal:
I moved the code using `Ssl` where it now belongs.  Note that I did not reused the `default_ctx` when a certificate is provided, because of the imperative nature of `Ssl.use_certificate`. This has the side effect that when using a password-protected certificate multiple times, a password is prompted for on each call. I think the correct solution is to create the context in `Conduit_lwt_unix.init` and embed it in the wrapper context.

paurkedal:
@hannesm Thanks for looking at it, and sorry for the amend-commits which I think interfere with the reviews, at least it makes it difficult to see what you commented on.

hannesm:
On 22/06/2017 06:53, Petter Urkedal wrote:
> I should have mentioned that I only tested with OpenSSL yesterday. The native fails with `Tls_lwt.Tls_alert(6)` there is no error code 6 in `Tls.Packet.alert_type`). But this is probably an issue for ocaml-tls, it's probaly a lack of overlapping ciphers.

Yes, this is some unfortunate serialisation (I tried to find out where
the 6 comes from, but to no success).  You can achieve better success by
manually setting up exception handlers (in case Lwt.async is used, use
the async_exception_hook):

Lwt.async_exception_hook := (function
  | Tls_lwt.Tls_alert a -> Printf.eprintf "tls alert: %s"
(Tls.Packet.alert_type_to_string a)
  | Tls_lwt.Tls_failure a -> Printf.eprintf "tls failure: %s"
(Tls.Engine.string_of_failure a)

paurkedal:
@hannesm That helped, well hook didn't work, but regular catching gave `HANDSHAKE_FAILURE`.  Which can be explained by the fact that we only support ECDHE ciphers on the site.

avsm:
@paurkedal apologies for the delay -- I'll merge this as soon as conduit.1.0 with the new build system is out.

paurkedal:
@avsm Sure, I noticed the activity.  Before you merge, I'll take another look whether I can integrate the certificate loading into the context without breaking the mold.

paurkedal:
Integrating certificate loading into the context seems difficult without creating a kludge or breaking backwards compatibility, which I'd attribute to lack of abstraction in `Conduit_lwt_unix`.

The main issue is this.  For the client side I would like to load the certificate data into `ctx` or `client` so that it can be re-used for multiple connections.  The `client` type cannot be changed in a backwards compatible way.  That leaves `ctx`.  This is doable, but leaves the library is a somewhat inconsistent state, since `ctx` is also used for the server, and the function `endp_to_server` provides a way to extract the certificate paths, rather than the data.  So, client and server side certificates would need to be handled differently, even though they use the same context.  We could have both the certificate paths and certificate data there, but it seems a inelegant, esp. since going though `endp_to_*` would cause reloading of certificates.

So, it's maybe better to load the certificate on connect for now, as the current PR does?

When 1.0 is ready, I can resolve the conflict.  Do you have a preference for how?  I noticed that a rebase-force-commit interferes with reviews.

vbmithr:
Ping. I also need Client TLS support :)
Rebased at http://github.com/vbmithr/ocaml-conduit#tls-client-auth.
@avsm @rgrinberg

avsm:
I've only just gotten to looking at this; thanks for bearing with me!  It seems better to me to put the client and server certificates in the ctx, and adapt the `endp` functions so that they understand better whether they are being used for client or server side authentication.

I don't quite understand your point about going through the endp functions causing reloading of certificates: can't the consumer of the function detect it's the same certificate and not do anything?

Either way, I've merged this into the next release with the current state so that the functionality exists -- we can do a rev of the interface in a later release again (with less latency this time, I hope!)

avsm:
Merged via #303

paurkedal:
Thanks!

I don't recall my precise reasoning, but looking at the code again, I think my point was to avoid loading the certificate on each `Conduit_lwt_unix.connect`, e.g. when interacting with a REST interface where a new connection is made per request.  (Yes, it could cache the certificate data.)

(About revision of the interface, I sketched out [an ml with more abstraction](https://gist.github.com/paurkedal/d41ec36b053b26f445927d954068cfce) around the time I made this PR.)

avsm:
That makes sense, thanks! @dinosaure has been prototyping out a modernisation of the overall conduit interface, so I'll factor in this PR into that discussion after ICFP is done :-)

PR #224, opened 2017-06-04
[wip] Jbuilder port

avsm:
This one requires some backwards incompatible changes, so I am opening this PR up early to get some feedback, cc @rgrinberg @samoht @djs55 @yomimono 

- Async_ssl is no longer an optional part of Conduit_async. There used to be portability issues, but nowadays I think it's ok to assume Async_ssl works.

- Going to split up the opam packages into `conduit`, `conduit-lwt`, `conduit-lwt-unix`, `conduit-mirage` and `conduit-async`.

- For `conduit-lwt-unix`, @rgrinberg suggested build time support for switching between Tls or OpenSSL. We could do this via two separate `opam` packages, to make the choice very explicit (e.g. `conduit-lwt-unix` and `conduit-lwt-unix-openssl`, with the default being ocaml-tls.

- `vchan` can be used by default in xen since its portable outside, as can `launchd` 

- `vchan_unix` is problematic as it does require some xen installation and so cannot be used by default i think. but it also has very few users -- can we just deprecate this for now @djs55?


yomimono:
I don't have a problem with the split scheme here, but considering the package universe breakage we've been seeing with some jbuilder ports I think it's reasonable to suggest reviving `mirage-dev` or another custom remote for testing these changes.

avsm:
definitely. cstruct wasn't too bad but not perfect...

avsm:
ssl is now optional again using http://jbuilder.readthedocs.io/en/latest/jbuild.html?highlight=man#alternative-deps

avsm:
@yomimono I dont think we need a full remote for conduit as it's not as bad as cstruct for revdeps...

```
cohttp                 0.22.0  HTTP(S) library for Lwt, Async and Mirage
datakit                0.10.1  Orchestrate applications using a Git-like dataflow
datakit-bridge-github  0.10.1  A bi-directional bridge between the GitHub API and Datakit
datakit-ci             0.10.1  Continuous Integration service using DataKit
datakit-github          0.8.1  A bi-directional bridge between the GitHub API and Datakit
git                     1.4.3  Git format and protocol pure OCaml
git-unix               1.11.0  Unix backend for the Git protocol(s)
hiredis                   0.2  Redis tools based on the Hiredis C library
jitsu                   0.3.0  A forwarding DNS server that automatically starts unikernels on demand
ketrew                  3.1.0  A Workflow Engine for Computational Experiments
mirage-conduit          2.3.1  Virtual package for the MirageOS Conduit transports
mirage-http             3.1.0  MirageOS-compatible implementation of the Cohttp interfaces
websocket                 2.6  Websocket library
```

@rgrinberg -- we could just release the new conduit and cohttp together and upper bound everything that currently uses them both

avsm:
reported the async_ssl build failures upstream https://github.com/janestreet/async_ssl/issues/25

rgrinberg:
> Async_ssl is no longer an optional part of Conduit_async. There used to be portability issues, but nowadays I think it's ok to assume Async_ssl works.

I think this will have negative consequences. I've certainly experienced many recent build failures with Async_ssl, so I disagree with the assessment that it just works. Using async without ssl is also not unheard of - when writing a server for example.

> we could just release the new conduit and cohttp together and upper bound everything that currently uses them both

Sure. I will make a cohttp release as soon as this conduit is released. Note that I will release the new cohttp once jsoo 3.0 is released. So there might be 2 releases there.

avsm:
Ok, I will make async_ssl optional as well then

avsm:
i think this might be almost ready for merge to master... just need to decide what to do about Vchan_lwt_unix support. A quick look in opam shows no users for this particular mode, so I am tempted to pull support for it and reintegrate it in a future release with more tests/etc.  Does anyone in xapi use Vchan_lwt_unix @jonludlam @mseri?

djs55:
I think the vchan Unix userspace support is fairly niche -- it's difficult to use because it doesn't fit the standard socket client/server model and relies on Linux user space grant and event channel interfaces (in particular the ability to transmit a grant as needed by a frontend was not used by anyone else in user space and was very buggy last time I tried). I think we can pull support for it and, in the unlikely event that a user is found, they can either add a version constraint or we can put the support back.

mseri:
It's fine with us. We don't currently use it or need it. In the worst case future scenario we can help re-instating the support or move it to an external package.

avsm:
everything passes now except for the async ssl failure which is being handled upstream. Could i get a few lgtm to merge, and then we can start with cohttp/conduit releases and revdeps.

rgrinberg: can we decouple the cohttp-js release in a post-cohttp.3.0.0 release? I suspect it has very few users and they can use an older cohttp for the short term...

rgrinberg:
Sure, but nevertheless let me release a version of cohttp that works against the old conduit first. This is so that we don't force users to upgrade both conduit and cohttp at once.

avsm:
do you think it would be better just to encourage a batch upgrade of both?  We only have to do the revdep work once then.  E.g. if cohttp-async were to depend on conduit-async, that simplifies everything a lot.

lib/jbuild
```diff
@@ -0,0 +1,8 @@
+(library
+ ((name        conduit)
+  (public_name conduit)
+  (wrapped false)
+  (preprocess (pps (ppx_sexp_conv)))
+  (modules (conduit conduit_trie resolver))
```
rgrinberg: Probably not the right time for this, but we should probably break compat at some point and set wrapped to true here. resolver is quit a common name to squat.

lwt-unix/jbuild
```diff
@@ -0,0 +1,20 @@
+(library
+ ((name        conduit_lwt_unix)
+  (public_name conduit-lwt-unix)
+  (preprocess (pps (ppx_sexp_conv)))
+  (wrapped false)
+  (optional)
```
rgrinberg: optional? Doesn't this have its own opam package?

avsm: oops that is fixed in 7a6784dfd8a68e57e741b3687058b72452cbbdc5

lib/jbuild
avsm: yeah, I have a rewrite in mind anyway with some of the lessons from httpaf and the new multicore IO work... I suspect that its not worth minor interface breakage, but just head towards a conduit2 that is blazing fast for connection establishment :-)

rgrinberg:
As long as conduit doesn't introduce any breakage, I think it's fine to couple the upgrade. Unfortunately, we usually find out about breakage after making a release. So if you're confident that such a scenario can be avoided, then please go ahead with it.

avsm:
Yeah -- I think if we cut some candidate releases, try some revdeps in opam-repository, but not merge until we're confident, it should be ok. I suspect that the combination of different conduit+cohttp versions interacting will be worse than just ripping off the depopt bandaid in one go...

lwt-unix/jbuild
```diff
@@ -0,0 +1,19 @@
...
+  (public_name conduit-lwt-unix)
+  (preprocess (pps (ppx_sexp_conv)))
+  (wrapped false)
+  (modules (resolver_lwt_unix conduit_lwt_unix conduit_lwt_server conduit_lwt_tls conduit_lwt_unix_ssl conduit_lwt_launchd))
+  (libraries (conduit-lwt tls uri.services ipaddr.unix logs 
+    (select conduit_lwt_launchd.ml from
+     (launchd  -> conduit_lwt_launchd_real.ml)
+     (!launchd -> conduit_lwt_launchd_dummy.ml))
+    (select conduit_lwt_unix_ssl.ml from
+     (lwt.ssl  -> conduit_lwt_unix_ssl_real.ml)
```
hcarty: Should these be `lwt-ssl` to match the new naming?

avsm: Good point - we can depend on the Lwt 3 series with this release so yes

Makefile
```diff
@@ -1,77 +1,7 @@
...
-install:
-	rm -rf _install
-	mkdir -p _install
-ifneq ("$(wildcard _build/lib/conduit_xenstore.cmo)","")
-	echo '"scripts/xenstore-conduit-init" {"xenstore-conduit-init"}' > _install/bin
-endif
-	$(foreach f,$(FILES), echo "$(f)" >> _install/lib;)
-	ocamlfind remove conduit || true
-	ocamlfind install conduit META $(FILES) $(MORE_FILES)
+	jbuilder build
```
samoht: please add `-dev` here

Makefile
```diff
@@ -1,77 +1,7 @@
...
-
-VERSION = $(shell cat VERSION)
-NAME    = conduit
-ARCHIVE = https://github.com/mirage/ocaml-$(NAME)/archive/v$(VERSION).tar.gz
-
-release:
-	git tag -a v$(VERSION) -m "Version $(VERSION)."
-	git push upstream v$(VERSION)
-	$(MAKE) pr
-
```
samoht: can you also add the multi-package release script as in https://github.com/mirage/irmin/blob/master/Makefile#L14-L24 Not perfect but that will help to make future releases.

Makefile
```diff
@@ -1,77 +1,7 @@
...
-# have(openssl) -> generate test certificates
-# have(lwt.unix tls.lwt ipaddr.unix lwt.ssl openssl) -> build(cdtest) and run(cdtest)
-# have(lwt.unix tls.lwt ipaddr.unix openssl) -> build(cdtest_tls) and run(cdtest_tlsOB)
-test:
-	$(OCAMLBUILD) tests.otarget
-	! openssl version || tests/unix/gen.sh
-	! ocamlfind query lwt.unix tls.lwt ipaddr.unix lwt.ssl || ! openssl version || ($(MAKE) cdtest && ./cdtest.native)
-	! ocamlfind query lwt.unix tls.lwt ipaddr.unix || ! openssl version || ($(MAKE) cdtest_tls && ./cdtest_tls.native)
-	! ocamlfind query lwt.unix tls.lwt ipaddr.unix || ! openssl version || (CONDUIT_TLS=native ./cdtest_tls.native)
+	rm -rf _build *.install
```
samoht: Can you also re-add a `test` target?

```
test:
    jbuilder runtest --dev
```

lwt/jbuild
```diff
@@ -0,0 +1,8 @@
+(library
+ ((name        conduit_lwt)
+  (public_name conduit-lwt)
+  (preprocess (pps (ppx_sexp_conv)))
+  (wrapped false)
+  (modules (resolver_lwt))
+  (libraries (conduit lwt))))
+(jbuild_version 1)
```
samoht: can you put the `(jbuild_version 1)` stanza at the beginning of the `jbuild` file? It applies to all `jbuild` files :-)

mirage-conduit.opam
```diff
@@ -7,19 +7,18 @@ bug-reports:  "https://github.com/mirage/ocaml-conduit/issues"
 tags:         "org:mirage"
 license:      "ISC"
 
-build:   ["ocamlfind" "query" "conduit.mirage"]
+build: [
+  ["jbuilder" "subst" "-p" name "--name" name] {pinned}
```
samoht: Not sure you need to repeat `-p` and `-n` here. `["jbuilder" "subst" "-n" name]` should be enough.

pkg/pkg.ml
```diff
@@ -0,0 +1,8 @@
+#use "topfind"
+#require "topkg-jbuilder"
```
samoht: please use `#require "topkg-jbuilder.auto` and remove the rest of the file.

samoht:
Thanks for the patches, I welcome build simplifications for this package, especially removal of pre-processing and optional dependencies. I've let a few (minor) comments.

Would be even better to also fix the API with extensible types, but I guess that will be for `conduit2` :-)

pkg/pkg.ml
avsm: cant do that, since `mirage-conduit` reverses the order, and I want to maintain backwards compat with the old packages (for now)

Makefile
avsm: done in f07d3dcd3e108c1b74415577b489de7f193a106d without `--dev` since there are still known warnings (due to interface change with lwt3, not fixed these yet)

Makefile
avsm: sure -- any chance you could add this to the mirage.io topkg instructions? very useful

avsm: done in 790d14f5cefc7c9b8b1982978422e562631bfdbd

lwt/jbuild
avsm: done in 9207e619d33a15be87b04a4a739d660a9be8c789.

@diml does `jbuild_version` really have to be the top line? it would be nice if jbuilder would warn if it appeared in an incorrect place if that's so

Makefile
avsm: no that breaks the build as there are known warnings. will add in a future release

lwt-unix/jbuild
avsm: hm, leaving this alone for the moment, since the depopts have to remain `ssl` for compatibility.  In a future revision we can move the whole lot to use `lwt-ssl` consistently, as there will be no further depopts left except for the sole use of `ssl`

lwt/jbuild
diml: @avsm it can be anywhere. I guess if one day we change the lexer then files using the new version will have to put the version first. For now it doesn't matter

Makefile
samoht: yea, updating the mirage.io instructions is on my todo list, I'll try to do this before the end of the week

Makefile
samoht: can't you add `[@@ocaml.warning "-42"]` in the code instead?

rgrinberg:
Excellent. The errors on travis are spurious.

PR #226, opened 2017-06-21
jbuild fixes

vbmithr:


lwt-unix/jbuild
```diff
@@ -5,14 +5,14 @@
...
-  (libraries (conduit-lwt tls uri.services ipaddr.unix logs 
+  (libraries (conduit-lwt tls.lwt uri.services ipaddr.unix logs
     (select conduit_lwt_launchd.ml from
      (launchd  -> conduit_lwt_launchd_real.ml)
      (!launchd -> conduit_lwt_launchd_dummy.ml))
     (select conduit_lwt_unix_ssl.ml from
-     (lwt.ssl  -> conduit_lwt_unix_ssl_real.ml)
-     (!lwt.ssl -> conduit_lwt_unix_ssl_dummy.ml))
+     (lwt_ssl  -> conduit_lwt_unix_ssl_real.ml)
+     (!lwt_ssl -> conduit_lwt_unix_ssl_dummy.ml))
```
avsm: thanks! I believe this requires bumping the minimum lwt version to >=3.0.0 to account for the new package structure?

hcarty: It should work with 2.7.0 or later - see https://github.com/ocsigen/lwt/blob/master/CHANGES

avsm: ah yes, thanks!

PR #232, opened 2017-07-25
Fix mirage-conduit ocamlfind dependencies

samoht:
Otherwise, we get:

```
File "_none_", line 1:
Error: No implementations provided for the following modules:
         Tls_mirage referenced from /Users/thomas/.opam/vpnkit/lib/mirage-conduit/conduit_mirage.cmxa(Conduit_mirage)
```

when linking

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,9 +5,9 @@ environment:
   CYG_ROOT: "C:\\cygwin"
   CYG_BASH: "%CYG_ROOT%\\bin\\bash -lc"
   TESTS: "false"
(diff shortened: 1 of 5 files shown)


avsm:
Bah, we need a test case for this as well here -- is there a simple linkage test that would find this in the future, without pulling in the whole `mirage configure` dance?

samoht:
@avsm: yea, just need to build some tests ;p I can add a very simple one, at least to check that linking is working fine.

samoht:
Simple linking test pushed

PR #235, opened 2017-08-03
Fix build dependencies

samoht:
Not specifying these could cause some weird races, where the ppx drivers are
removed while the library is being built.

See https://travis-ci.org/mirage/irmin/jobs/260197274 for an example

### DIFF
--- a/conduit-lwt-unix.opam
+++ b/conduit-lwt-unix.opam
@@ -14,6 +14,7 @@ build: [
 depends: [
   "base-unix"
   "jbuilder" {build & >="1.0+beta9"}
+  "ppx_sexp_conv" {build}
   "conduit-lwt"
   "lwt"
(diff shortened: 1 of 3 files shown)


samoht:
Change submitted in opam repo as well: https://github.com/ocaml/opam-repository/pull/10011

samoht:
This fixed the weird (but reproducible) error seen in https://github.com/mirage/irmin/pull/467

rgrinberg:
Hmm, I think cohttp might suffer from the same problem. I still don't understand what's the cause here though. Why would ppx_drivers be removed while the lib is being built? They're a transitive requirement still.

samoht:
> They're a transitive requirement still.

Where? This could happen when you are reinstalling packages, as opam will do that in parallel when possible.

PR #239, opened 2017-09-05
Remove warnings

samoht:


lwt-unix/conduit_lwt_unix_ssl_real.ml
```diff
@@ -61,7 +60,7 @@ module Server = struct
       ?timeout sa cb =
     sa
     |> listen ~ctx ?backlog ?password ~certfile ~keyfile
-    |> Conduit_lwt_server.init ?stop (fun (fd, _) ->
+    >>= Conduit_lwt_server.init ?stop (fun (fd, _) ->
```
samoht: I don't understand how conduit-ssl was compiling without that line. Are we testing this properly?

rgrinberg: Looking at travis, it seems like we don't test this variant with lwt ssl enabled. We should though

samoht:
Don't merge this yet, I've extracted the SSL fix in #241 where I've added more travis tests. I will rebase this PR once #241 is merged.

PR #241, opened 2017-09-06
Fix compilation of the ssl backend

samoht:
Also complete the matrix in the travis script so that future regressions will be caught faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,16 +7,14 @@ script: bash ./.travis-docker.sh
 env:
  global:
    - PINS="conduit:. mirage-conduit:. conduit-async:. conduit-lwt:. conduit-lwt-unix:."
-   - PACKAGE="conduit"
-   - DEPOPTS="mirage-conduit conduit-async conduit-lwt conduit-lwt-unix"
  matrix:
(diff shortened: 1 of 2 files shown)


rgrinberg:
Is this a complete matrix though? We should also test the dummy modules i.e. whenever ssl or async_ssl are absent but conduit-lwt-unix or conduit-async are built.

samoht:
yes, it's what `DEPOPTS` is supposed to do: first build without extra-package, then install the depopts. I'll double check that's it's indeed case.

samoht:
So it's compiling both with and without the all the depopts. So I should probably try to alternate between tls/ssl/both. Also it seems that the tests are not run (very simple tests atm, just trying to link stuff together) so I am trying to fix that.

samoht:
Tests are not running because we need a new release of ocaml-ci-scripts, which is now in progress: https://github.com/ocaml/ocaml-ci-scripts/pull/179

samoht:
So it appears that fixing the CI is much more involved that I was planning. Given that the current released version of conduit is broken, I'll focus now on making a new release - and I (or someone else) could fix the CI later.

Issue #244, opened 2017-11-07
endpoints should keep all routes


rixed:
`system_resolver` takes the first result from getaddrinfo and discard all others. But when several routes exist to reach a hostname all might not work. In practice, this issue arises often when a server listen only to IPv4 or only to IPv6 and the server hostname is resolved into both an IPv4 and an IPv6 addresses.
conduit should keep all routes for an endpoint and try them all in turn when connecting.

Currently, it is for instance not possible to start a cohttp server on localhost (will bind only IPv4) and connect to it using cohttp client, if the system resolves to IPv6 first (which is the case on my Debian stable box and on my MacOs laptop).

vbmithr:
Hitting the same issue on Tezos. By default, the system resolver returns IPv6 first (on my machine), and since conduit reverse this, `localhost` does mean IPv6 for Tezos node and IPv4 for Tezos client.

vouillon:
See [RFC 8305 (Happy Eyeballs Version 2: Better Connectivity Using Concurrency)](https://tools.ietf.org/html/rfc8305)

dinosaure:
This issue is _partially_ solved when this is the responsibility of the user to rightly use `getaddrinfo` now and give a function to `conduit` to explain how to correctly resolve a `[ host ] Domain_name.t`:
```ocaml
let my_resolver domain_name = match Unix.getaddrinfo (Domain_name.to_string domain_name) "tcp" [] with
  | _ :: _ as lst ->
    let[@warning "-8"] best :: _ = my_sort lst in
    Lwt.return_some best.Unix.ai_addr
  | [] -> Lwt.return_none
  | exception _exn -> Lwt.return_none

let resolvers =
  Conduit.empty
  |> Conduit.add Conduit_lwt.TCP.protocol my_resolver

let run domain_name =
  Conduit.resolve ~resolvers domain_name >>= fun flow ->
```

However, the RFC referenced by @vouillon highlighted for us some possible mis-conception on the API level. You if have any feedbacks about that, feel free to open an issue with a knowledge of the new API of `conduit`. Thanks for your report.

PR #245, opened 2017-11-07
Favor resolving into IPv4 over IPv6

rixed:
If the server is listening to both it would make little difference.  If the
server is listening only to IPv4, as does cohttp, then it makes the endpoint
usable. If the server is listening only to IPv6, as does no HTTP server I know
about, then the endpoint is now broken.

A better fix would be to keep all addresses and try them all in turn, as
suggested in #244.


avsm:
thanks! We'll take a look at #244 in a future release.

PR #248, opened 2017-11-21
fix compilation on async dev

vbmithr:
Needed to compile against `async`, dev version.

### DIFF
--- a/async/conduit_async.ml
+++ b/async/conduit_async.ml
@@ -15,6 +15,7 @@
  *
  *)
 
+open Core
 open Async
 
 module Ssl = Conduit_async_ssl.Ssl_config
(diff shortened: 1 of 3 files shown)


rgrinberg:
Does this require us to tweak the upper bounds for async?

vbmithr:
If you want to compile against the dev version of `async`, you need this. It depends on the policy of this repo. If this repo tracks the stable branch of async (I guess it is the case), then this will be merged only when Jane Street release a new version of `async`. Otherwise, if this repo tracks the bleeding-edge branch of `async`, this can be merged now, but indeed there is a lower bound to add on `async`.

rgrinberg:
We definitely track the latest stable release of async. But I mean if this PR should be updated with an appropriate upper bound, so that it is ready to merge when the new async is released.

avsm:
there will be a new release of async incoming to opam-repo early next week, cc @trefis

trefis:
Yes there was a backward incompatible change in the dev version of async recently (sorry for not going through the usual "deprecated for n-months then removed" cycle), so once you merge this you'll only be compatible with the latest version of async (which will probably be v0.10) and incompatible with v0.9.

So yes, this PR should also change the opam file.

copy:
async v0.10.0 is now in the opam repository and has been marked as conflicting with the current version of conduit-async in the process (https://github.com/ocaml/opam-repository/pull/11130), so this should be tested and merged now.

rgrinberg:
Thanks. I cherry picked this to master and fixed the upper bounds. Also undid some of the style changes. The less opens the better...

vbmithr:
Which commit? I don't see anything on master.

copy:
> Which commit? I don't see anything on master.

It's not on master, but in the `pr/248` branch. `opam pin add conduit-async 'https://github.com/mirage/ocaml-conduit.git#pr/248'` works for me.

rgrinberg:
Sorry, I pushed it to the correct branch now.

PR #251, opened 2017-12-31
compile against async_extra v0.9.116.03+91

bmillwood:
I'm using [janestreet/opam-repository](/janestreet/opam-repository) and needed to make these changes to compile against it. I guess this shouldn't be merged until the next JS package release hits opam, but this is (a subset of) what will be necessary when that happens :)

### DIFF
--- a/async/conduit_async.ml
+++ b/async/conduit_async.ml
@@ -33,22 +33,22 @@ type addr = [
 let connect ?interrupt dst =
   match dst with
   | `TCP (ip, port) -> begin
-      Tcp.connect ?interrupt (Tcp.to_host_and_port (Ipaddr.to_string ip) port)
+      Tcp.connect ?interrupt (Tcp.Where_to_connect.of_host_and_port { host = Ipaddr.to_string ip; port })
       >>= fun (_, rd, wr) -> return (rd,wr)
   end
   | `OpenSSL (_, ip, port) -> begin
-      Tcp.connect ?interrupt (Tcp.to_host_and_port (Ipaddr.to_string ip) port)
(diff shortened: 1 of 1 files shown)


async/conduit_async.ml
```diff
@@ -110,7 +110,7 @@ type server = [
 
 let serve
       ?max_connections ?backlog
-      ?buffer_age_limit ?on_handler_error mode where_to_listen handle_request =
+      ?buffer_age_limit ?(on_handler_error=`Ignore) mode where_to_listen handle_request =
```
bmillwood: I didn't actually think too hard about whether this is the correct default. The options are:

    -> on_handler_error       : [ `Raise
                                | `Ignore
                                | `Call of ('address -> exn -> unit)
                                ]

The default used to be `` `Raise``. I vaguely recall that this turned out to be a bad default because it meant that any random process connecting to your port and talking the wrong protocol might take down your server, so it's easy to DoS you even accidentally. But `` `Ignore`` is clearly also kind of a bad default because it's easy to ignore errors. Perhaps the right thing to do is mirror `Tcp.Server.create` and make the argument mandatory. But mostly I just wanted to get it to compile for my own project, and didn't personally fancy fixing any downstream clients :)

bmillwood: relevant async_extra commit: bb08ddacc1ab6c413a58ffbd8667292d2cd82e26

I guess this should be `` `Raise`` so that it's at least behaviour-preserving. I'll go make that change.

rgrinberg:
Thanks, I think @vbmithr already made a PR for this before and it's now in master. Let us know if anything else needs to be done.

bmillwood:
Ah, bad timing on my part. Thanks :)

PR #255, opened 2018-03-06
Implement SNI (Server Name Indication) for SSL backend

vouillon:
This addresses mirage/ocaml-cohttp#378.


rgrinberg:
Can we make the hostname optional to avoid the breaking change?

vouillon:
The hostname is now optional. I did not realized that `Conduit_lwt_unix_ssl.Client.connect` was part of the public API.

rgrinberg:
Thanks.

PR #256, opened 2018-03-09
Conduit (wrapped)

rgrinberg:
This PR adds `(wrapped true)` to `conduit` and `conduit-lwt`. I think both of these changes can be introduced with minimal breakages. I don't believe that anyone is using names like `Revolver`, `Conduit_trie`, or `Resolver_lwt` directly.


yomimono:
`Resolver_lwt` is used directly by Mirage when constructing unikernels using conduit for name resolution: https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L1078 .

rgrinberg:
Thanks for noticing this. We can coordinate that change in mirage, right?

yomimono:
Yes.  Conveniently I've just written up a first stab at the list of actions required to make changes like this.  I'll post that somewhere slightly more open and we can see whether the instructions are complete witht his change.

lib/core.mli
```diff
@@ -44,7 +44,7 @@
 
     All of the name resolvers conform to the {!RESOLVER} module type.
     The OS-specific implementations of this interface are:
-    {!modules: Resolver_lwt Resolver_lwt_unix Resolver_mirage}
+    {!modules: Conduit_lwt.Resolver Conduit_lwt.Resolver_unix Resolver_mirage}
```
samoht: I don't think `Resolver_lwt_unix` has been renamed by this PR

lib/intro.html
```diff
@@ -29,7 +29,7 @@
 
 All of the name resolvers conform to the {!Resolver.S} module type.
 The OS-specific implementations of this interface are: 
-{!modules: Resolver_lwt Resolver_lwt_unix Resolver_mirage}
+{!modules: Conduit_lwt.Resolver Conduit_lwt.Resolver_unix Resolver_mirage}
```
samoht: I don't think Resolver_lwt_unix has been renamed by this PR

lib/core.ml
```diff
@@ -0,0 +1,35 @@
+(*
```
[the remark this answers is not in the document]
rgrinberg: OK, sure. I've used core before in cases where actual code was present. I agree that `S` is more appropriate. Btw, should we make the names in `Conduit.S` the recommended ones and deprecated the `Conduit.foo` names.

rgrinberg:
@yomimono ping on this. Would be nice to have this for 1.2.0

yomimono:
The doc is at http://canopy.mirage.io/Posts/Deprecating , although @aantron had some very good edits that aren't reflected there.

If you expect 1.2.0 to be some time from now, we might first release a version of `mirage` which adds an upper bound on `conduit-lwt` in the generated opam files.

In advance of 1.2.0, we'll need to change the code in `mirage` along with the version specifications for the `conduit-lwt` package (and probably add a conflict on earlier versions of `conduit-lwt` for the `mirage` front-end tool), then release that version of `mirage` with the correct version bounds once version 1.2.0 is released.

yomimono:
Sorry, I dropped our end of this but I'm happy to include the constraints necessary in a 3.1.1 if you're still willing to push from the conduit side.

rgrinberg:
Sure, the work is pretty much done. Ping me when you want me to cleanup the pr and merge.

avsm:
We're going to rework conduit quite majorly with @dinosaure 's work, so not going to do this PR I think.

PR #257, opened 2018-03-20
Fix file descriptor leakage on EADDRINUSE

rixed:
No idea why it hasn't been done before :-/


avsm:
lgtm (I had to double-check `with_socket` to make sure it didnt finalise the fd, but it purely catches errors and closes it instead)

rgrinberg:
Errors in CI are spurious. LGTM

PR #261, opened 2018-04-18
Force users to give a custom callback in case of exception to avoid random crashes

kit-ty-kate:
When using Conduit_lwt_unix.serve with cohttp, for instance, in a very simple http server that serves files, the server crashes with an exception when the connexion is cut while the server is trying to send the response. For example this happens very frequently if you are spamming Ctrl-R on your browser.

This happens because the ```on_exn``` parameter defaults to ```print an error message with a backtrace if available and to exit the program``` (quote from the ```Lwt.async_exception_hook``` documentation).

My proposal is simple: it is really easy to make this mistake, the error is quite difficult to understand and might happens at any moment, and therefore the user should be forced to choose what to do in case an exception is raised.

cc @rgrinberg

### DIFF
--- a/lwt-unix/conduit_lwt_unix.ml
+++ b/lwt-unix/conduit_lwt_unix.ml
@@ -318,7 +318,7 @@ let serve_with_default_tls ?timeout ?stop ~ctx ~certfile ~keyfile
   | No_tls -> failwith "No SSL or TLS support compiled into Conduit"
 
 let serve ?backlog ?timeout ?stop
-    ?(on_exn=(fun exn -> !Lwt.async_exception_hook exn))
+    ~on_exn
     ~(ctx:ctx) ~(mode:server) callback =
   let callback flow ic oc =
     Lwt.catch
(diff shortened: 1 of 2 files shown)


mseri:
Can you also send a PR to update https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-lwt-unix/src/server.ml#L56 and the corresponding `mli` (to be merged in parallel with this and released after this)? In there is maybe worth to add an example, e.g.

```ocaml
let on_exn = function
  | Unix.Unix_error (error, func, arg) ->
    Logs.warn (fun m ->
      m  "Client connection error %s: %s(%S)"
        (Unix.error_message error) func arg
    )
  | exn -> Logs.err (fun m -> m "Unhandled exception: %a" Fmt.exn exn)
)
```

just to point out what it could look like, or maybe even add something like that as the default, to make the behaviour slightly saner

mseri:
What's the status of this? I think this is a welcome change to have

PR #267, opened 2018-07-16
Do not link with tls.lwt on windows

### LINKED ISSUE #264: Linking, tls and windows
It seems we have a _linking bug_ when we want to have the SSL support on `cohttp`. I really don't know what happens but, on `ocaml-git`, if I want to use `tls`, I did not have any problem on Linux but (cf. mirleft/ocaml-nocrypto#135) I have a problem on Windows.

So, a solution could be "(ssl | tls)" as a dependency of `git-unix` but [neither Travis nor Appveyor is happy](https://github.com/mirage/ocaml-git/pull/277) ... Then, with a little discussion with @samoht, the problem seems to be in `conduit`. So it's why I created this issue.

samoht:
Fix #264

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,11 @@ env:
    - PINS="conduit:. mirage-conduit:. conduit-async:. conduit-lwt:. conduit-lwt-unix:."
    - TESTS=true
  matrix:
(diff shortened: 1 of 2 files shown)


samoht:
note: I haven't tested the PR at all yet

dinosaure:
`async_ssl.0.11.0` does not compile.

PR #275, opened 2018-10-05
Use dune

samoht:
/cc @emillon and @hannesm


lib/dune
```diff
@@ -0,0 +1,7 @@
+(library
+  (name        conduit)
```
emillon: you can drop the `(name)` and bump the version to 1.1 (lang & opam). If that's your thing :)

mirage-conduit.opam
```diff
@@ -8,13 +8,13 @@ tags:         "org:mirage"
 license:      "ISC"
 
 build: [
-  ["jbuilder" "subst" "-n" name] {pinned}
-  ["jbuilder" "build" "-p" name "-j" jobs]
+  ["dune" "subst" "-n" name] {pinned}
```
emillon: I'm not sure `dune subst` supports `-n`. Anyway it's safe to remove since it's in `dune-project`.

samoht:
Any idea why ounit can't install on debian-unstable?

emillon:
curl failed, it looks just transient. I'd retry the build.

samoht:
still no luck :-(

samoht:
Merging it anyway, as it seems to be an infrastructure failure.

edwintorok:
I think the curl failure of ounit on debian unstable is because the version of openssl there drops support for TLS1.0 and the forge doesn't support it. There is a https://forge-static.ocamlcore.org/ mirror that would work.

PR #276, opened 2018-10-05
Mirage types

samoht:
On top of #275 

Remove the dependency to `mirage-types` and use more precise types instead.


hannesm:
I reviewed the second commit, looks good to me.

Issue #188, opened 2016-12-30
conduit.mirage no longer being installed


juanchanco:
The 0.14.* series seems to no longer install conduit.mirage. As a result, installing mirage-conduit (via opam) fails. Is it no longer needed? Currently, mirage-conduit is a dependency for mirage-http.

samoht:
My bad, I've merged changes in master which should have been merged with the `mirage-dev` branch instead. I am currently fixing this.

juanchanco:
Everything works find with version 0.14.2 now, thanks.

PR #277, opened 2018-11-01
Get client IP when using TLS/SSH connections

victorgomes:
This is related to issue #188 
The client IP and port wasn't properly propagated.


victorgomes:
I've been using this patch in our server: https://cerberus.cl.cam.ac.uk

deuszex:
I was directed here from issue number #286
To run the pr had to downgrade ipaddr to 2.9.0 and tls to 0.9.2 then it worked as expected and got the ip of the client who connected in to our server

deuszex:
will there be anything of this?

PR #282, opened 2018-12-15
add server type (`Listening_socket of Lwt_unix.file_descr)

### LINKED ISSUE #281: Support systemd socket activation, binding a server to an open Unix.file_descr
I'm using cohttp_lwt_unix to make a server, which takes a `mode` pf type `Conduit_lwt_unix.server`. I'd like to use systemd socket activation (ideally via https://github.com/juergenhoetzel/ocaml-systemd/), which requires binding to an already-open `Unix.file_descr`, which doesn't appear to be supported by the existing type: https://github.com/mirage/ocaml-conduit/blob/a030ed47209ff35464c43f7d07bef21fccc9569e/lwt-unix/conduit_lwt_unix.ml#L83-L92

I see there's already a `Launchd socket activation option, so I'm assuming it'd be a similar code path.

timbertson:
Fixes #281 . The variant had to be of type `sexp_opaque` because `file_descr` is abstract. That would cause a runtime error trying to serialize one of these, but I assume it won't break other variants.

(network buffs: is "listening socket" the right description for an inherited systemd socket? AFAIK you don't need to call `listen` but you do need to call `accept` on such a socket, so it sounds right to me)

### DIFF
--- a/lwt-unix/conduit_lwt_unix.ml
+++ b/lwt-unix/conduit_lwt_unix.ml
@@ -89,6 +89,7 @@ type server = [
   | `Vchan_direct of int * string
   | `Vchan_domain_socket of string  * string
   | `Launchd of string
+  | `Listening_socket of Lwt_unix.file_descr sexp_opaque
 ] [@@deriving sexp]
 
 type tls_server_key = [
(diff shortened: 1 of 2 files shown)


avsm:
This looks right to me -- just requesting another review from @djs55 who wrote the launchd backend

timbertson:
bump?

avsm:
Merging, thanks!

PR #287, opened 2019-02-07
Addr of uri

vbmithr:
This adds a convenience function to resolve `Uri.t` into an `addr` as used by `conduit-async`.

### DIFF
--- a/async/conduit_async.ml
+++ b/async/conduit_async.ml
@@ -17,6 +17,7 @@
 
 module V1 = V1
 module V2 = V2
+module V3 = V3
 
 [@@@deprecated "Use Conduit_async.V1"]
 include V1.Conduit_async
--- a/async/dune
+++ b/async/dune
@@ -1,10 +1,10 @@
 (library
  (name        conduit_async)
  (public_name conduit-async)
- (modules     conduit_async private_ssl v1 v2 s)
- (preprocess  (pps ppx_sexp_conv))
+ (modules     conduit_async private_ssl v1 v2 v3 s)
+ (preprocess  (pps ppx_here ppx_sexp_conv))
  (libraries
-   conduit async
+   conduit async ipaddr.unix uri.services
 
     (select private_ssl.ml from
      (async_ssl  -> private_ssl_real.ml)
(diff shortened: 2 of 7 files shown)


vbmithr:
@rgrinberg @seliopou

avsm:
This is more than a convenience function though: it totally changes the type of `addr`.  What's the motivation for that?

vbmithr:
To be able to return the underlying socket as Async permits.

avsm:
Ok... and are there any upstream users? How bad is the compat break, and is there some way to test the downstream consumer of the new change?

vbmithr:
I don't know other users. I'm modifying interface V2, that we designed with @rgrinberg to be able to integrate changes without breaking other users. Not sure what you mean by testing the downstream consumers, but it's just an interface change so functionality is unchanged.

avsm:
I basically have no context for this PR, so I'm trying to understand what it's for, who will be using it, and what the impact of the changes will be.

vbmithr:
1. What it's for:
1. Convenience function to resolve an URL
2. Get the underlying Socket.t when using connect/with_connection

2. Who will be using it: me, at least

3. Impact. Interface change for V2, which we designed to be the "unstable" interface for Conduit_async.

avsm:
Thanks for this information. But there's nothing really that indicates the `V2` is an unstable API (in 898f849ff330f8905c25fcbd8afe6f17c7bf35f8 at least), and it's been in two released versions of conduit-async.  Given we have versioning, why not make this new improved interface a `v3.ml` and add deprecation annotations to V2?  If it's only you using it, there should be very little additional burden, but it does make transitioning other users more explicit.

vbmithr:
Ok seems reasonable. I hope I won't need to change things every couple of weeks, but I don't see why either, since this is quite simple (though already the third iteration!)

emillon:
I think it would be helpful to leave V3 marked as unstable for a few releases before deprecating V2. This will make it possible to add new features or change the V3 API for a little while.

avsm:
This looks good to merge; will do after #294

PR #288, opened 2019-02-20
Fix exception raised when using Async_ssl when other side disconnects.

bogdan2412:
One needs to be careful around Async Readers and Writers that share the
same underyling file descriptor, which is something that happens when
they're used for sockets.

Closing the Reader before the Writer will cause the Writer to throw and
complain about its underlying file descriptor being closed.

This happens when using Async_ssl bindings because Reader.pipe closes
the underlying Reader once the Reader reaches EOF. This then makes
the Writer sad because it shares the same socket file descriptor.

Instead of using Reader.pipe, we write out our own variant which will
close the Writer before closing the Reader.

### DIFF
--- a/async/private_ssl_real.ml
+++ b/async/private_ssl_real.ml
@@ -8,6 +8,23 @@ let verify_certificate connection =
   | Some (Error _) -> return false
   | Some (Ok _) -> return true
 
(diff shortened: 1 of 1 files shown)


bogdan2412:
Moving pull request to its own branch.

Issue #231, opened 2017-07-25
rename mirage-conduit into conduit-mirage


samoht:
Not sure why this is in reverse vs. all other `conduit-*` package names.

avsm:
because this is what it was before -- I wanted to leave the package names as similar as possible and then make all the changes in one go across various projects (e.g. mirage-http, mirage-conduit, mirage-dns) and to clearly deprecate the old packages to avoid confusion.  I'm still confused by tar-format->tar and git-mirage, since the older packages are still around.

samoht:
I think removing the names which doesn't follow the convention would remove the confusion :-)

But yes, I guess it's ok to do that progressively!

avsm:
but we can just delete the old packages -- they are still in opam-repository and show up in searches. So we do need some mechanism (e.g. `messages`) of deprecating them and/or edit the short descriptions to reflect that they were part of MirageOS2 or something...

avsm:
for example try `opam list --depends-on mirage-unix` and try to figure out which of those packages are still relevant :)

samoht:
Duplicate of #178

PR #290, opened 2019-03-19
use dns-client >= 4.0.0 in mirage-conduit

hannesm:
replaces the dns resolver implementation. while at it, I found some code which I could safely remove (I tested end-to-end with canopy). see https://github.com/roburio/udns/pull/14 for changes needed in udns

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,6 +1,0 @@
-PKG ppx_driver.ocamlbuild
-PKG sexplib stringext uri cstruct ipaddr lwt vchan async_ssl ssl tls tls.lwt
-PKG dns.mirage mirage-types mirage logs
-B _build/**
-S lib/
-S tests/**
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
    - PINS="conduit:. mirage-conduit:. conduit-async:. conduit-lwt:. conduit-lwt-unix:."
    - TESTS=true
  matrix:
+   - OCAML_VERSION=4.07 PACKAGE=mirage-conduit   DISTRO=alpine
    - OCAML_VERSION=4.07 PACKAGE=conduit-lwt-unix DISTRO=debian-stable   DEPOPTS="ssl tls"
    - OCAML_VERSION=4.06 PACKAGE=conduit-lwt-unix DISTRO=debian-unstable DEPOPTS="ssl tls"
    - OCAML_VERSION=4.05 PACKAGE=conduit-async    DISTRO=debian-unstable DEPOPTS=async_ssl
--- a/mirage-conduit.opam
+++ b/mirage-conduit.opam
@@ -28,3 +28,4 @@ build: [
   ["dune" "runtest" "-p" name] {with-test}
 ]
 dev-repo: "git+https://github.com/mirage/ocaml-conduit.git"
+synopsis: "A network connection establishment library for MirageOS"
--- a/mirage-conduit.opam
+++ b/mirage-conduit.opam
@@ -14,6 +14,7 @@ depends: [
   "mirage-stack-lwt" {>= "1.3.0"}
   "mirage-time-lwt" {>= "1.1.0"}
   "mirage-flow-lwt" {>= "1.2.0"}
+  "mirage-random" {>= "1.2.0"}
   "dns-client" {>= "4.0.0"}
   "conduit-lwt"
   "vchan" {>= "3.0.0"}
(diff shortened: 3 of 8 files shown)


hannesm:
//cc @dinosaure @cfcs turns out most of resolver_mirage is not used (i tried cohttp-mirage, irmin-mirage and git-mirage as revdeps), and was safe to remove :D

hannesm:
oh, and there's actually a resolver API at https://roburio.github.io/udns/doc/udns-mirage-client/Udns_mirage_client/Make/index.html FWIW

dinosaure:
I noticed these behaviors about `tuyau` where `ocaml-uri` is less paranoid but I prefer it (which mostly follow RFC) than before. My only concern is about the default resolver where `92.239.100.100` is a part of `udns`'s internals and, from the patch, it's not clear that we use it now.

I don't know if it's possible to let end-user to specify a default resolver in `udns` and notice `92.239.100.100` statically here.

hannesm:
@dinosaure in [Mirage_impl_resolver](https://github.com/mirage/mirage/blob/v3.5.0/lib/mirage_impl_resolver.ml#L43-L46) we already support `~ns ~ns_port` labeled optional arguments, which are threaded to Udns. since the udns client is the library that needs a nameserver, I'm ok with it having a hardcoded (well, actually better would be multiple) recursive resolver.

hcarty:
What does `91.239.100.100` map to?

hannesm:
@hcarty anycast.censurfridns.dk, see https://blog.uncensoreddns.org/

hcarty:
@hannesm Thanks!

hannesm:
I especially would like to know whether the `module type S` (and the `Make` functor), removed in this PR, is used by someone somewhere (I tried reverse dependencies of mirage-conduit -- namely cohttp-mirage, git-mirage, and irmin-mirage -- they all don't use it)!? (it was introduced 4 years ago in https://github.com/mirage/ocaml-conduit/commit/f16721ce58c573864a6d8b7ae3026e54e7cf955b, but to me it looks like nobody is using the plain `Make` anymore, and nobody ever used `S`)!?

avsm:
The only one i can think of that might be using it is https://github.com/moby/vpnkit, cc @djs55

djs55:
I had a quick look -- there's no use of `conduit` in `vpnkit`. So feel free to delete `module type S` and `Make` with a clear conscience!

hannesm:
thanks! I still have to release udns, but this is likely to happen very soon now :)

avsm:
I am massively looking forward to switching away from the venerable ocaml-dns :-)

hannesm:
@avsm yes, i hope https://github.com/roburio/udns/compare/next is my last rewrite before an initial release

hannesm:
this PR has been revised to the revised API of dns-mirage-client (taking the random device as functor argument) -- also the ns and ns_port are now (https://github.com/mirage/mirage/pull/996) parameters (configuration time and runtime) -- thus I think all the issues have been addressed, and this is good to go and being released once dns is released (tomorrow!)

since I've not much clue about conduit releases, is it viable to merge, tag, and release only mirage-conduit (with version 4.0.0)? or how is the tooling to release conduit* as y.z and mirage-conduit as a different version (since they started off differently)!?

hannesm:
ok, this was a bit more involved than anticipated. I took the opportunity (since there's anyways a split universe between those who use dns-1.y and dns-4.y) to rename `mirage-conduit` to `conduit-mirage` here -- this allows us to reset the version numbers for `conduit-mirage` as well (and start e.g. at 1.5.1, where other conduit packages are at -- no more manual messing around for a release).

this in addition adds a travis job which compiles conduit-mirage. if this is acceptable, please say so (or merge and release directly -- conduit-mirage is not yet used by anyone). once this is in opam, I'm happy to adjust other packages (most of which, since the API / module have not changed, may even just need a `depends: ("mirage-conduit" | "conduit-mirage")`, i.e. work with either old or new package names! :)

as a side-effect this fixes #178 then!

hannesm:
the main argument from #231 is not valid anymore, since the dominant naming is now YY-mirage (as can be seen on the dependencies onto mirage-conduit http://opam.ocaml.org/packages/mirage-conduit/) for YY in {git,cohttp,irmin,resp,h2,dns}.

hannesm:
turns out, it is a bit more work than adjusting opam files, since the references in `dune` need to be adjusted as well -- I'm fine PRing these changes to the projects, pls let me know what you think about the rename. I can as well revert these commits from this PR, and we can rename another time.

hannesm:
force-pushed the mirage-conduit changes only, the rename is over in https://github.com/hannesm/ocaml-conduit/tree/rename -- feel free to PR that branch here and merge if you feel like...

PR #292, opened 2019-05-06
lwt-unix: use safe close when creating channels from fds

vbmithr:
Without this patch, closing both channels resulting from the same file descriptor will always raise an exception for the second close.

### DIFF
--- a/lwt-unix/conduit_lwt_unix.ml
+++ b/lwt-unix/conduit_lwt_unix.ml
@@ -160,6 +160,13 @@ let init ?src ?(tls_server_key=`None) () =
     | {ai_addr;_}::_ -> Lwt.return { src=Some ai_addr; tls_server_key }
     | [] -> Lwt.fail_with "Invalid conduit source address specified"
 
+let safe_close fd =
+  Lwt.catch
+    (fun () -> Lwt_unix.close fd)
+    (function
+      | Unix.Unix_error (EBADF, _, _) -> Lwt.return_unit
+      | exn -> Lwt.fail exn)
+
 (* Vanilla sockaddr connection *)
 module Sockaddr_client = struct
   let connect ?src sa =
@@ -168,8 +175,10 @@ module Sockaddr_client = struct
          | None -> Lwt.return_unit
          | Some src_sa -> Lwt_unix.bind fd src_sa) >>= fun () ->
         Lwt_unix.connect fd sa >>= fun () ->
-        let ic = Lwt_io.of_fd ~mode:Lwt_io.input fd in
-        let oc = Lwt_io.of_fd ~mode:Lwt_io.output fd in
+        let ic = Lwt_io.of_fd
+            ~close:(fun () -> safe_close fd) ~mode:Lwt_io.input fd in
+        let oc = Lwt_io.of_fd
+            ~close:(fun () -> safe_close fd) ~mode:Lwt_io.output fd in
         Lwt.return (fd, ic, oc)
       )
 end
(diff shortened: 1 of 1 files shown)


hcarty:
I haven't tested this yet, but good catch!  I'll try to test tomorrow as I have tons of log lines warning about these `EBADF` errors.

dinosaure:
Hmmhmm, I understand the purpose of this PR but I don't like to silently catch `EBADF` and return unit. I would like, at least, a warning to continue to keep tracking this kind of error. Otherwise, it seems that `conduit` has a bad design for this purpose, but it's an other problem.

vbmithr:
I need a concrete solution to avoid logging thousands of useful lines in my logs, as well as issuing thousands of bogus syscalls.

lwt-unix/conduit_lwt_unix.ml
```diff
@@ -160,6 +160,13 @@ let init ?src ?(tls_server_key=`None) () =
     | {ai_addr;_}::_ -> Lwt.return { src=Some ai_addr; tls_server_key }
     | [] -> Lwt.fail_with "Invalid conduit source address specified"
 
+let safe_close fd =
+  Lwt.catch
+    (fun () -> Lwt_unix.close fd)
```
hcarty: Can we check with [Lwt_unix.state](http://ocsigen.org/lwt/4.1.0/api/Lwt_unix#VALstate) here before the call the `Lwt_unix.close`?  Then we can avoid the double-close without having to ignore the `EBADF`.

Something like (typed directly into github without testing):
```ocaml
let close_if_open fd =
  match Lwt_unix.state with
  | Lwt_unix.Aborted _
  | Lwt_unix.Opened -> Lwt_unix.close fd
  | Lwt_unix.Closed -> Lwt.return_unit
```

hcarty:
@dinosaure Would https://github.com/mirage/ocaml-conduit/pull/292#r282625786 be acceptable to avoid the double-close without swallowing all possible `EBADF` cases?  If so, please feel free to use it in this PR @vbmithr

dinosaure:
Yes I prefer this idea where we ensure that `BADF` came from it. Then, a notice on documentation will be perfect :+1: !

avsm:
I’ve been digging into Lwt_io and it really seems like a regression, since channels are supposed to be lazily double closed without issue. Not sure what the root cause of this is

hcarty:
@avsm In this case, though, it's two channels which share an underlying descriptor.  Would that guarantee still hold?

avsm:
Ah! It's not safe to have two channels share the same fd then, I suspect.  The problem with catching the EBADF and ignoring it is that there is a potential race that is even worse:

- Channel A and B share fd 1
- Channel A closes fd 1
- New fd is created that reuses 1 
- Channel B closed fd 1 -- which is the new fd, and no EBADF is raised but a random other fd is closed

Given that Lwt_io channels do the close at most once, we seem to have to ensure that the channels themselves dont share fds to prevent double closes like the above

vbmithr:
This is a bug in Lwt's API then. Should open a bug report there.

hcarty:
I had a chat with @avsm and @dinosaure.  The issue is that we have two channels backed by the same file descriptor without handling the double-`close` appropriately.  It could be considered a bug in the Lwt API but I'm not sure it's an implementation bug.  May be worth clarifying the Lwt documentation.

I have a branch using `Lwt_io.open_connection`, based on @avsm's suggestion.  I'll push a PR once I've done a bit more testing.  It seems to address the double-close issue.

avsm:
@vbmithr could you test #294 to see if it causes any other regressions in your setup?

vbmithr:
Ok sure, will do.

avsm:
Will be fixed via #294, thanks for the report!

vbmithr:
Working for me, thanks! Closing this, superseded by #294

PR #295, opened 2019-06-22
Async_ssl backend bug fixes

bogdan2412:
- Fix exception raised when using Async_ssl when other side disconnects.

One needs to be careful around Async Readers and Writers that share the
same underyling file descriptor, which is something that happens when
they're used for sockets.

Closing the Reader before the Writer will cause the Writer to throw and
complain about its underlying file descriptor being closed.

This happens when using Async_ssl bindings because Reader.pipe closes
the underlying Reader once the Reader reaches EOF. This then makes
the Writer sad because it shares the same socket file descriptor.

Instead of using Reader.pipe, we write out our own variant which will
close the Writer before closing the Reader.

- Add missing teardown logic for Reader.of_pipe and Writer.pipe.

[Writer.pipe] will not close the writer if the pipe is closed, so in order to avoid
leaking file descriptors, we allow the pipe 30 seconds to flush before closing the
writer.

[Reader.of_pipe] will not close the pipe when the returned [Reader] is closed, so we
manually do that ourselves.

[Writer.of_pipe] will create a writer that will raise once the pipe is closed, so we
set [raise_when_consumer_leaves] to false.

### DIFF
--- a/async/private_ssl_real.ml
+++ b/async/private_ssl_real.ml
@@ -8,6 +8,23 @@ let verify_certificate connection =
   | Some (Error _) -> return false
   | Some (Ok _) -> return true
 
+(* One needs to be careful around Async Readers and Writers that share the same underyling
+   file descriptor, which is something that happens when they're used for sockets.
+
+   Closing the Reader before the Writer will cause the Writer to throw and complain about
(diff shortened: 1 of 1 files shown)


avsm:
Good catch; very similar to the Lwt one also fixed in #294

PR #296, opened 2019-06-27
[WIP] ssh client support in conduit_mirage

hannesm:
please do not merge as is, this is on top of #290 (in this PR, only 5f1f30f is relevant)

the abstraction provided by conduit (`client`), loosely speaking: given an Uri.t provide a flow (with read/write/close) is not sufficient for ssh (FWIW, same is true for tls):
- ssh needs (well, certainly you can use a password if you like) a private key for authentication, as well as an authenticator (usually a server fingerprint, but maybe as well a CA certificate), and a command to execute (this is not talking about the fully general ssh protocol, just the bits needed for git+ssh)
- tls should accept an authenticator as well, also an optional client certificate

it is not clear to me how this fits / should fit into conduit:
- atm i used an optional `string`(`?config`) and serialise/deserialise the command, private key seed, and authenticator -- this should be slightly better typed (sth like a ssh config)
- unclear whether conduit should depend on the Tls.Config.client and Awa.Client.client directly or provide other structures

--> it is as well a bit unclear whether the goal of conduit is to be extensible in terms of transport protocols (i.e. should there be a conduit-mirage-tls, conduit-mirage-ssh)?

### DIFF
--- a/conduit-mirage.opam
+++ b/conduit-mirage.opam
@@ -25,6 +25,7 @@ depends: [
   "tls-mirage" {>= "0.11.0"}
   "ipaddr" {>= "3.0.0"}
   "ipaddr-sexp"
+  "awa-mirage"
 ]
 conflicts: [
   "mirage-conduit"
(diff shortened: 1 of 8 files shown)


hannesm:
sorry, i pressed enter to quickly, and just revised the description above.

dinosaure:
Fixed by #311.

PR #298, opened 2019-07-22
Expose verify_mode in Conduit_async

brendanlong:
Previously the verify_mode was private, which made it impossible
to turn on certificate verification.


brendanlong:
I also found as a workaround, you can set `~verify:Conduit_async.Ssl.verify_certificate`

avsm:
Thanks!

PR #305, opened 2019-10-25
conduit-mirage: do not raise in DNS lookups

hannesm:
previously, e.g. a http cient connecting to http://foofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofooa/ would result in an exception (label length <= 63, here 64).

previously, a http client connecting to http://foo_bar would result in an exception (hostname only consists of letters digits and hyphen (which may not be the first letter))


hannesm:
since this only affects the conduit-mirage, travis is happy with it, and there are some more outstanding changes (that depend/conflict with this), I'll merge. please raise your voice if you'd like it to be reverted. ;)

Issue #308, opened 2019-12-10
Unix HTTP client "resolution failed: unknown scheme" if /etc/services is missing


mseri:
This issue was raised in the ocaml-cohttp repo but I think it may be worth to crosspost it here. See here for details: https://github.com/mirage/ocaml-cohttp/issues/675

dinosaure:
Due to the fact that the new version of `conduit` does not use `/etc/services` anymore, I think this issue can be closed - however, it highlight a possible regression which is already discussed here: https://github.com/mirage/ocaml-conduit/pull/311#issuecomment-655594905

mseri:
This should be reopened, it is currently showing as an issue in cohttp-lwt-unix when tests run in nix

Issue #286, opened 2019-01-15
Getting client IP address using endp_of_flow


deuszex:
Greetings!
I seem to have somewhat the same problem as one that was already fixed in issue #87 
A code as follows will always return 0.0.0.0 as the ip address, while if i replace the ~src with my ip, than that will be the ip returned at every incoming connection. If i don't give ~src the result will be again 0.0.0.0

```
let start_server () =
  Conduit_lwt_unix.init ~src:"0.0.0.0" () >>= fun ctx ->
    Conduit_lwt_unix.serve ~ctx ~mode:(`TLS config) begin fun flow ic oc ->
      let addr = match Conduit_lwt_unix.endp_of_flow flow with
      | `TCP (ip, port) ->
        (Ipaddr.to_string ip) ^ ":" ^ (string_of_int port)
      | _ -> "" in
      Lwt_io.printf "incomming connection from %s\n" addr;
    end
;;
```

Isn't endp_of_flow supposed to return the source ip of the incoming call, or did I make a mistake somewhere?

Issue appears on both 1.3.0 and 1.4.0 versions.

hannesm:
There is an open pull request about client IP address - https://github.com/mirage/ocaml-conduit/pull/277 - could you try that PR and see whether it solves your issue?

deuszex:
I just tried that PR does not seem to work for us

deuszex:
i stand corrected, it does work when you downgrade the ipaddr package to 2.9.0 and the tls package to 0.9.2

dinosaure:
According to the new version of `conduit` and this comment, we are able now to _destruct_ the given `Conduit.flow` to the underlying `Lwt_unix.file_descr` - and get the IP of the client. See the comment: https://github.com/mirage/ocaml-conduit/issues/119#issuecomment-701386235

The code to get the IP address of the client is:
```ocaml
Conduit_lwt.resolve ~resolvers domain_name >>= function
| Ok (Conduit_lwt.TCP.T lwt_socket) ->
  let peer = Lwt_unix.getpeername lwt_socket in
  ...
| _ -> ...

Issue #280, opened 2018-11-16
Do not close connection after read


Silur:
Now the situation reversed from https://github.com/mirage/ocaml-conduit/issues/274
```ocaml
let start () =
  Conduit_lwt_unix.init ~src:"127.0.0.1" () >>= fun ctx ->
  let serve () =
    Conduit_lwt_unix.serve ~ctx ~mode:(`TLS config) begin fun addr ic oc ->
      Lwt_log_core.notice "Server: Callback started." >>= fun () ->
      Lwt_io.read ~count:255 ic >>= fun msg ->
      Lwt_io.printf "got data %s" msg
    end
  in
  serve ()
;;
```
now netcatting into the server ALWAYS results in a server close
```bash
openssl s_client -connect 127.0.0.1:19979
foobar
closed
```

SGrondin:
I can confirm this. Adding an unnecessary 100ms sleep on the client side before sending data 'fixes' it.

dinosaure:
So I tried to retranslate your code with the new version of `conduit`, the highlighted bug is a bit hard to track over the major release  and I'm not sure which behavior you want. So with this code:
```ocaml
open Lwt.Infix

let () = Mirage_crypto_rng_unix.initialize ()

let load_file filename =
  let ic = open_in filename in
  let ln = in_channel_length ic in
  let rs = Bytes.create ln in
  really_input ic rs 0 ln ;
  close_in ic ;
  Cstruct.of_bytes rs

let config cert key =
  let cert = load_file cert in
  let key = load_file key in
  match
    (X509.Certificate.decode_pem_multiple cert, X509.Private_key.decode_pem key)
  with
  | Ok certs, Ok (`RSA key) ->
    Tls.Config.server ~certificates:(`Single (certs, key)) ()
  | _ -> invalid_arg "Invalid key or certificate"

let tls_config = config "server.pem" "server.key"

let tcp_config =
  { Conduit_lwt.TCP.sockaddr= Unix.(ADDR_INET (inet_addr_loopback, 19979))
  ; capacity= 40 }

let handler flow =
  let flow = Conduit_lwt.pack Conduit_lwt_tls.TCP.protocol flow in
  let ic, oc = Conduit_lwt.io_of_flow flow in
  Lwt_log_core.notice "Server: Callback started." >>= fun () ->
  Lwt_io.read ~count:255 ic >>= fun msg ->
  Lwt_io.printf "got data %s" msg

let fiber =
[...]

Issue #273, opened 2018-09-13
No way to stop a server that hasn't started yet


rixed:
I have a server that fail in a loop because the TCP port is already bound.
I'm trying to stop this server, either by:
1. `Lwt.cancel` the thread  (actually the thread returned by Cohttp.Server.create);
2. `raise Exit` from the `on_exn` callback;
3. calling the `stop` thread.
to no avail.

Is there any way to interrupt a server in this state?

dinosaure:
Not sure to really understand your issue but if a port is already used by someone else, the new version of `conduit` will just fail to initialize the _server_. I mean, if you do:

```ocaml
let tcp_config =
  { Conduit_lwt.TCP.sockaddr= Unix.(ADDR_INET (inet_addr_of_loopback, 8080))
  ; capacity= 40 }
let run =
  Conduit_lwt.Service.init tcp_config ~service:Conduit_lwt.TCP.service >>= function
  | Ok socket -> ...
  | Error err ->...
```

You will get an `Error (`Msg "Address 127.0.0.1 already in use")`, see this part of `conduit-lwt`:
https://github.com/mirage/ocaml-conduit/blob/66fd34a9dc1dc2507f6c20752e589ae53cc4ccf8/src/lwt/conduit_lwt.ml#L156-L192

So, I consider that your issue is fixed by the new version of `conduit`. If you have any troubles, feel free to re-open the issue. Thanks!

rixed:
That's likely the case that a more recent version of conduit fixes this bug, indeed.

PR #311, opened 2020-04-20
3.0.0

dinosaure:
This is the main PR where I replaced all conduit by `tuyau`. Currently, the `lwt-unix` implementation is missing and should be provided as before. Otherwise, it seems good to prepare a plan about a possible release of 3.0.0.


dinosaure:
It seems that the CI is happy with my new test: a ping-pong server without and with TLS. I did not yet implement the `async` support. Only the `lwt-unix` support is available - about `mirage`, I don't know the way to compile/run properly a `mirage` _unikernel_.

I currently on a reverse-dependency work to smoothly move to the new interface.

dinosaure:
Considering this _TODO_ before any merge:
- [ ] cohttp
- [ ] datakit
- [ ] git
- [ ] hiredis
- [ ] jitsu
- [ ] ketrew
- [ ] websocket
- [ ] yurt

nickbetteridge:
I would find cohttp, git and websocket highly useful :)

- datakit has been archived

samoht:
> datakit has been archived

it's not anymore.

nickbetteridge:
Ahah, then datakit would be a welcome inclusion also! :)

lib/conduit.ml
```diff
@@ -52,7 +52,7 @@ module type S = sig
   val send : flow -> output -> (int, [> `Msg of string ]) result s
   val close : flow -> (unit, [> `Msg of string ]) result s
 
-  type 'edn resolver = [ `host ] Domain_name.t -> ('edn option) s
+  type 'edn resolver = [ `raw ] Domain_name.t -> ('edn option) s
```
hannesm, line 55: I've a question about this commit: did you encounter any non-host domain names that are used by conduit clients? IMHO it would be nice to use `` [ `host  ] Domain_name.t``.

dinosaure: Yes, I'm not fixed about that and before, I used `[ host ]`. However, it seems that into `ocaml-cohttp`, or precisely into `Uri.t`, we can pass an IPv4 such as `http://0.0.0.0/` which does not contain a valid `[ host ] Domain_name.t`.

A possible solution is to update the `cohttp`'s test suite to pass `localhost` instead.

hannesm: if an Uri can be like this, it is also special in conduit/tuyau, since there's no need to resolve it via DNS -- i.e. that step can be saved (and when doing TLS, it should also verify the IP address in the certificate rather than the hostname [this is not implemented IIRC])....

but i think it is worth keeping host and handling ip addresses in a uri.t special (of course this may be a separate PR)

dinosaure:
TCP `async` support was done to be able to provide `cohttp-async` as usual. The design is the same than `conduit-mirage.tcp` where we need to unlock the internal loop provided by `Async.Tcp.Server` to be able to provide `accept` as expected.

dinosaure:
We re-implement the support of SSL on `conduit-async`. I did not test the code but I did it as usual. However, we should add some tests about it:
- [ ] test `conduit-async` and `conduit-async.ssl`

conduit-tls.opam
```diff
@@ -0,0 +1,47 @@
...
+description: """
+The `conduit` library takes care of establishing and listening for 
+TCP and SSL/TLS connections for the Lwt and Async libraries.
+
+The reason this library exists is to provide a degree of abstraction
+from the precise SSL library used, since there are a variety of ways
+to bind to a library (e.g. the C FFI, or the Ctypes library), as well
+as well as which library is used (just OpenSSL for now).
+
+By default, OpenSSL is used as the preferred connection library, but
```
avsm, line 36: Just discussing this with @dinosaure  -- I think this paragraph is no longer true

lib/README.md
```diff
@@ -0,0 +1,127 @@
+## Conduit - core library
```
avsm, line 1: Could we move this library into the ocamldoc for `conduit`, so it shows up in the docs.mirage.io and others? Conduit 2 used an index.mld file for this

conduit-tls.opam
edwintorok: Will OpenSSL still be available as a backend? I think that is an important use-case to preserve, to let the user choose which TLS implementation to use.

dinosaure: Yes, I'm currently on that after a discussion with @avsm and we found a solution to avoid any regression on this specific purpose :+1:

dinosaure:
Support of `lwt_ssl` was done and it was tested. Currently, we already try `TLSv1_2`, it seems that in GitHub Actions (and on my computer), we don't have the support of `TLSv1_3`.

dinosaure:
For more context and about produced implementation of protocols, I stress-test `conduit` with `httpaf` to ensure any details.

hannesm:
what happened to vchan? I'm not sure whether vchan support is used by anyone (or needed - and for sure can be re-added later if desired).

dinosaure:
I finished to support SSL and TLS on `lwt` and `async` with examples and tests.
- [ ] windows support

dinosaure:
Ok, `ocaml-ci` is happy (including `ocamlformat`), I still have some troubles about tests only on `ocaml-ci` but I can not introspect it as a black-box. I removed all optional dependencies, so `conduit-async` and `conduit-lwt-unix` have an explosion of dependencies due to TLS and SSL layers provided.

`conduit` still is light (and `conduit-lwt`) and `conduit-mirage` needs `conduit-tls` to provide a TLS layer on top of the TCP/IP stack.

talex5:
> I still have some troubles about tests only on ocaml-ci but I can not introspect it as a black-box.

What is the problem? Each platform's log page displays a shell script you can run to reproduce the build on your local machine.

dinosaure:
> What is the problem? Each platform's log page displays a shell script you can run to reproduce the build on your local machine.

The problem is a bit hard to highlight. Currently tests want to start several _ping-pong_ servers and clients (with several protocols, TCP, TCP + SSL and TCP + TLS) and it seems that _sometimes_, `ocaml-ci` stuck on one of them (but logs can not tell me more about which one and why).

I added `(locks ...)` to serialize tests and avoid an EADDRINUSE but I don't really what happens for 116a42d, 2d615e5 or 64ef587.

dinosaure:
`cohttp` is done, tests are good, CI is green, let's go to others projects.

.github/workflows/test.yml
```diff
@@ -7,22 +7,27 @@ jobs:
     strategy:
       matrix:
         ocaml-version: [ '4.08.1', '4.09.0', '4.10.0' ]
-        operating-system: [macos-latest, ubuntu-latest, windows-latest]
```
avsm, line 10: why remove Windows from this? All the packages should compile on Cygwin too (since mirage-crypto got ported)

avsm, line 10: why remove Windows from this? All the packages should compile on Cygwin too (since mirage-crypto got ported)

dinosaure: Yes, I added a step about `windows`, it should works now when I fixed plenty of bugs. I'm currently on a rebase of this PR.

dinosaure:
The PR is cleaned. The support of Windows is a bit tricky. We currently can build and install `conduit-lwt-unix` (`conduit-async` can not be installed due to `core`/`async_ssl`) but tests fails because a out of scope error (`ssleay32.dll` not found).

I consider the PR as done and ready for reviews. `cohttp` works on top of that. I will re-evaluate the list of _revdeps_ projects because some of them are olds.

dinosaure:
- `datakit`
The project is a bit too hold to be updated with this version of `conduit`. It did not follow mirage/mirage#1004 (where `conduit` requires `mirage-flow.2.0.1`) and I don't have enough time to do a regular maintenance on it and update it to `conduit.3.0.0`
- `git` still is on the list when I maintain it
- `hiredis` it seems that the last version of it does not use `conduit` anymore (so, probably a constraint on the released one should be enough)
- `jitsu` requires the VCHAN support, I will deal with it but outside the scope of `conduit` (due to your extensible variant, we are able to do that now, let's take this opportunity.
- `ketrew` is mostly about `cohttp` where they start an HTTP server with TLS or SSL or without secure layer. At the end, I think it's a wrong design and the choice of the layer (or if we can use such layer) should not be done by a `depopts` game but must be chosen by the end-user (cc @hammerlab)
- `websocket` is a protocol, so it should be _easy_ to use the last version of `conduit`
- `yurt` wants to start a TLS server, and the choice to have a TLS server or not seems to be done by the end-user so an update with `conduit.3.0.0` is possible.

Then, I think that `ocaml-cohttp`  is _a drop in the ocean_. So an update of `conduit` will break the world definitely but the PR about `cohttp` is ready (and its CI is happy) so let's continue the upgrade!

avsm:
- jitsu can be considered deprecated. We can bring back vchan at a future date.
- datakit doesnt work with irmin 2.0, so also deprecated with conduit 3
- what about `irmin` itself? is that just dependent on cohttp?

zshipko:
Awesome, I’m happy to update `yurt` for conduit 3.0, but `hiredis` is more or less deprecated at this point

dinosaure:
> what about irmin itself? is that just dependent on cohttp?

It seems that complexity/real use of `conduit` is reported to `git` and `cohttp` if I look precisely into the code, we obviously need to update it but the patch should be really small!

dinosaure:
So all non-deprecated packages with a direct dependency to `conduit` (eg. `opam list --depends-on conduit`) have a PR to pave the way to the migration. It's a first iteration and a second one should appear after a release of `conduit.3.0.0` - `git` will be a part of this second iteration.

`irmin` will be a part of the second iteration too. The most important, I think, it's about `cohttp`.

dinosaure:
Finally, a conclusion is needed:
- [x] `conduit-async` and `conduit-lwt-unix` have a support of `openssl` available on `conduit-async.ssl` and `conduit-lwt-unix.ssl`. These packages are well tested (Linux and MacOS) with a simple ping-pong server.
- [x] into details, `conduit-async.ssl` provides as before a `Reader.t` and a `Writer.t` to be easy to use
- [x] `conduit-async` provides a way to create `Reader.t` and `Writer.t` from an abstract `flow`
- [x] into details, `conduit-lwt-unix` provides a way to create `Lwt_io.channel` as before (required by `cohttp`)
- [x] the main loop to create a _server_ is still available but it's not systematically used (to let the user to do something more fancy) with `serve_with_handler` (however, a regression exists when it does not handle `timeout` or `backlog` as before)
- [x] a `mirage-flow` of the abstracted type `flow` is provided by `conduit-lwt` and fallback by `conduit-mirage` as before - however, such implementation is deprecated now (it was well documented)
- [x] `conduit-lwt-unix` and `conduit-mirage` provides a TLS layer with `ocaml-tls`
- [x] we still use `[ host ] Domain_name.t` instead `Uri.t`:
  I think that _metadata_ given by an `Uri.t` should be handle by the end-user - or how to destruct an `Uri.t`, meaning of scheme if he wants to use, etc. At least, `conduit` needs a concrete type to do the resolution and I still think that `[ host ] Domain_name.t` is the best.
- [x] `conduit-async` and `conduit-lwt-unix` provides a way to create a `resolver` according the system (a convenience function to use `gethostbyname` - `resolv_conf`)
- [x] a document exists to explain how to use `conduit`
- [x] a simple example is done (`ping_pong.exe`) to show how to use `conduit`

Pro:
+ `conduit-async` provides a TLS layer with `ocaml-tls` now!
+ ability to register a protocol outside `conduit`
+ ability to use your own type _endpoint_ to initialize a connection
+ ability to compose a layer such as TLS (but SSL too) with another protocol
+ ability to concretely _extract_ the `flow` from a given definition

Cons:
* `vchan` protocol does not exist anymore. However, `vchan` package can provide by itself the required _endpoint_ and _protocol_ witness to be usable with new version of `conduit`
* Due to the fact that we use `[ host ] Domain_name.t`, we can not pass an IP to the resolution process - however, this case mostly appear to initiate a server which is now outside the scope of the resolution.

So, I will merge this PR this friday if nobody has a strong opinion about this update :+1: - `cohttp` will follow and we will be ready for the second iteration of updates.

samoht:
> I will merge this PR this friday 

Please wait a bit more, I'd like to review that PR but won't have time until next week.

dinosaure:
First, it seems that the last version of `conduit` import an unrelated conflict. I will fix it next week. Secondly, it seems that `conduit` has plenty of issues which can be resolved with this new version. This is a big summary about all them and in which way this new version solves issues.

#### TLS and `ocaml-tls`

#31, #147, #153, #155, #253 and #266 highlight the same problem:
- how we can _inject_ our TLS configuration
- how we can _project_ the used TLS configuration by the given `flow`

The first point seems really important where we currently don't have the ability to give the `authenticator` when `conduit` makes the `Tls.Config.client` - we can only give `certificates`. Of course, `Tls.Config.client` has several options and all of them have a importance magnitude depending on the context of the end-user. So, it should be a bit crazy to fallback all of these options to `conduit` - specially when `conduit` tries to find a common API between `ocaml-tls` and `libssl`.

The new version is more abstracted (with _type witness_) on this part. A `flow` can require a `Tls.Config.client` to start a secure connection with a peer. But the new version of `conduit` does not take the responsibility to create this value. Instead, it asks to the end-user to fill a local _heterogeneous map_ with some `resolvers`. One of them can give to `conduit` a special `Tls.Config.client` with specific options (such as a specific _authenticator_, some ciphers, etc.). Then, depending on which protocol is registered, `conduit` is able to pass this value to the `connect` function. Of course, always depending on your registered protocols, you can give something more complex.

In other side, if you know (in your dependencies) that it's possible to have a TLS flow, the new version of `conduit` has the ability to _destruct_ the `type flow` to the structural definition of it. Such function is:

```ocaml
val is : flow -> 'flow Witness.protocol -> 'flow option
```

From that, you are able to introspect the given configuration if you have a `(Tls.Config.client * 'flow) Witness.protocol`. Usually, this value should be provided by the implementer of the protocol. Luckily, `conduit-tls` gives to you such value by composition with another protocol.

#### Project the _real_ flow

#119, #155, #159 and #286 concern about a way to get the real value used into your given `flow`. This question is a bit hard because it splits into 2 cases the real use-case of `conduit`. In fact:
- in one side, we want a connection (without a full knowledge of which kind of connection we have)
- in other side, we want to talk with a specific protocol (and we want to fully-know the given `flow`)

In the first case, only `conduit-{lwt,async}` is enough and __as a library__ such as `irmin` or `ocaml-git` (where we must take care about dependencies), we just want to start a connection without a knowledge of which protocol we use.

By this way, we can avoid an hard dependency to something which implements a protocol such as:
- `mirage-tcpip`
- `lwt.unix` to get the TCP/IP UNIX stack
- `async` to get the TCP/IP UNIX stack

Depending only on what the end-user filled into the local _heterogenous map_, we will be able, in our library, to communicate with a peer. Of course, such use-case is really interesting for `mirage` where we can imagine `irmin` communicates with a peer on a serial-port - without a knowledge of how to communicate with a serial-port. 

Initially, `conduit` is done for that: to give an abstraction between libraries such as `irmin` and protocols such as TCP/IP.

However, the second feature is too important when we want to make an _executable_ or something less abstracted than `irmin`. Such case appears most of the time. So, as we said, `Conduit.is` exists but the ability to _project_ the real value is done only if you depend on the expected protocol. In others words, you can _project_ a given `flow` to a `Lwt_unix.file_descr` __only if__ you depends on `lwt.unix` (and `conduit-lwt-unix`). It seems logic but I would like to be clear on this point - the projection is not magic.

#### Server loop

I did not speak about that a lot but, #107, #112, #114, #273 and #280 have the same problem, implementation of the server loop. In fact, `conduit` comes with its own implementation of the server loop (which does the `accept`, call the handler and _yield_ the scheduler). However, I think that such a loop should be on the responsibility of the end-user to really control such complex process (when we stop, time-out, some others concurrent processes, etc.).

The new version of `conduit` comes with a new interface to describe a [_service_](https://github.com/mirage/ocaml-conduit/pull/311/files#diff-db3454c643be04c6e60ca05e527d337eR80). But I did not try to fix all of these problems by deleting codes! `conduit-{lwt,async}` provides [`serve_with_handler`](https://github.com/mirage/ocaml-conduit/pull/311/files#diff-fe0df0773275a873c7479e6c5a1ec919R17) which does what the old version of `conduit` did - however, some options such as the _time-out_ or _backlog_ disappeared.

So it's like what the end-user wants. He can use our function or make by himself the main loop where the given API is pretty close to what UNIX provides.

However, a special case appears for `Async.Tcp` and `mirage-tcpip`. In fact, both want to implement their main loop. In case of `Async.Tcp`, `async` provides low-level functions such as `accept` and we use them to implement the TCP/IP service for `conduit-async`. For `mirage-tcpip`, the trick is a bit more complex when we must _split_ the loop implemented by `mirage-tcpip`. In that case, we use a shared queue between the handler specified to `mirage-tcpip` and our `accept` function. More details are available [here](https://github.com/mirage/ocaml-conduit/pull/311/files#diff-ba94010e01d9cc42080304c053bcb305R233).

#### Optional dependencies

#3, #93, #187, #237 and #266 exist because the compilation of `conduit` depends on what you have in your OPAM environment. However, mirage ecosystem has a strong knowledge about optional dependencies and we should not use it.

This new version does not have optional dependencies. The counter part of this is the _explosion_ of dependencies needed by `conduit-lwt-unix` and `conduit-async` (specially `libssl` and `ocaml-tls`). The choice was made to provide SSL (and TLS with `libssl`) and TLS protocols on such packages. `conduit-mirage` depends on only (obviously) `ocaml-tls`.

I think it's fair to start with this new dependency graph than to randomly get an exception to say that `conduit` does not support TLS.

#### Endpoints and resolution

#9, #131, #244, #270, #271 and #308 are about _endpoint_ and the resolution process. The problem can be described as is:
- Who does the resolution?

And if we can have a response, an other question appear: What the resolver should return?

The new version of `conduit` finally abstracts the _endpoint_. We said that the user can give a `Tls.Config.client` but he is able to give to `conduit` anything (as long as _anything_ is described by a protocol). The usual example is to pass a `Unix.sockaddr` to be able to start a connection then. So, the question about IPv6 for example should not exist anymore where the only condition is to have a protocol which handles IPv6.

Come back to the first question, the new version of `conduit` did the choice to remove any _default_ resolvers. The end-user __must__ specify which resolver he wants. He can use the resolver of his system or use `ocaml-dns` (and directly talk to root DNS servers).

The counterpart is: at the beginning, `conduit` is not able to resolve any domain-name.

Of course, `conduit-lwt-unix` and `conduit-async` come with a convenience function to easily _register_ the system resolver into the local _heterogeneous map_.
```ocaml
val resolv_conf : port:int -> Lwt_unix.sockaddr resolver
```

Finally, the choice was made to be more explicit than before. User should do some extra operations than the previous version of `conduit` but he will be more aware about what happens.

#### Conclusion

It still exists some others issues but they are a bit too precise to consider them as fixed with this new version. Then, I don't want to say that this PR is the best and it fixes any of our problems but I think it paves the way into a good end according all of these issues. I hope this summary clears about next stage of `conduit`. Bon week end.

dinosaure:
A huge patch was done to simplify the API from what @samoht did (eg. dinosaure/ocaml-conduit#1). So I will describe more precisely updates (even if I explain a bit changes into my commit 3c25b46).

### One type-witness

On the old design, end-user must play with 2 type witness values:
- `type 'a key` which represents the value required to _initialize_ a connection
- `type 'a Witness.protocol` which represents the value to send and receive something (eg. the _socket_)

With such design, the user must take care about link between these values - and if he did something wrong, an error `Invalid_key` was returned in few places of the code.

@samoht came with the idea to merge these values into one: `('edn, 'flow) protocol`

This PR does that. At the end, the registration of a protocol still be the same but it does not require a `'a key` where such type witness value is created and used internally.

### Split `conduit` into 2 parts, client and server

It should be done at the beginning but @samoht did that on his side. I split it in this PR too which give to us a better API than before. `conduit` provides 2 sub-modules:
- `Conduit.Client` which implements:
    + `recv`, `send`, `close` and `connect`
    + some others functions like the registration of a protocol or some _accessors_ such as `is`, `abstract`, `impl_of_protocol` and `impl_of_flow`
- `Conduit.Service` which implements
    + `serve`
    + the registration function and `impl` to get the service implementation from the given type witness value.

### Pattern-matching on the type `flow`

@samoht tried to integrate the nice idea to be able to _pattern-match_ on the type `flow`. The idea is to be able to write this kind of code:
```ocaml
match my_flow with
| Conduit.Value (TCP.T file_descr) -> ...
| Conduit.Value (TLS.T (file_descr, tls_context)) -> ...
| _ -> ...
```

To easily take the advantage of the underlying used value. So the type `flow` is now defined as: `type flow = private ..`. Because we use `private`, you still need to use `Client.register` to extend it. The knowledge of `TCP.T` or `TLS.T` can be done by the protocol implementer by a simple:
```ocaml
: sig
  val protocol : (sockaddr, socket) Client.protocol
  type t = (sockaddr, socket) Client.value
  type Conduit.flow += T of t
end = struct
  let protocol = Conduit.Client.register ...
  include (val Conduit.Client.repr protocol))
end
```

dinosaure:
New goals are:
- [ ] documentation
- [ ] fallback `Service.accept`/`Service.close` into `conduit`
- [ ] _functorize_ over `Scheduler` `ping_pong.exe` as @samoht did
- [ ] tests `resolvers`
- [ ] integrate tests from `mirage-flow` into `conduit`

dinosaure:
The branch is rebased and ready to merge.

tests/flow.ml
```diff
@@ -0,0 +1,219 @@
...
+  let pp ppf = function
+    | #Rresult.R.msg as v -> Rresult.R.pp_msg ppf v
+    | `Not_found -> Fmt.string ppf "`Not_found" in
+  let equal a b = match a, b with
+    | `Msg a, `Msg b -> a = b
+    | `Not_found, `Not_found -> true
+    | _ -> false in
+  Alcotest.testable pp equal
+
+module Memory_flow0 = struct
```
samoht, line 35: Can you add this in `Conduit` directly (the usual name for this is `Loopback`) ?

tests/flow.ml
```diff
@@ -0,0 +1,219 @@
...
+  let res4 = Conduit.recv flow Bytes.empty in
+  Alcotest.(check (result send error)) "res0" res0 (Ok 4) ;
+  Alcotest.(check (result send error)) "res1" res1 (Ok 4) ;
+  Alcotest.(check (result send error)) "res2" res2 (Ok 4) ;
+  Alcotest.(check (result send error)) "res3" res3 (Ok 0) ;
+  Alcotest.(check (result recv error)) "res4" res4 (Ok `End_of_flow) ;
+  Alcotest.(check string) "buf" (Bytes.to_string buf) "Hello World!" ;
+;;
+
+module Memory_flow1 = struct
```
samoht, line 123: What's the difference with `Memory_flow0` ?

dinosaure: `Memory_flow0` computes a `string` as an input and a `bytes` as an output.
`Memory_flow1` computes a `string list` as an input and a `bytes list` as an output.

The second can be the first if we `List.concat`.

tests/flow.ml
dinosaure: I'm not sure because:
- such "protocol" works with `string` as an input and `bytes` as an output - so it requires an application of `Conduit.Make` which differs from the used application with `cstruct` by `conduit-lwt` and `conduit-async` (see l. 8). At the end, it will not be possible to compose it with others `conduit-*`.

Note that I tried to abstract such types into witnesses but it reveals something more complex then.

- for me, such impl. is only relevant about tests. In this view, we should not expose it.

samoht: Well, having an efficient in-memory flow to connect various process is very useful in some situation (not always for tests) ; it would be a shame if every user would have to write one. 

I agree that maybe Memory_flow0 or Memory_flow1 are not the best candidates for these as they do not work on cstructs as the rest of the client library.

dinosaure: I see your point, I can try to provide something with `Cstruct.t` as `conduit-tls` does.

samoht:
After a first glance, the changes looks ok, but I need to have another look.

Can you also split the libraries into:
- conduit (the core functor)
- conduit-lwt (a core subpackage, but the main unix package and the namespace should be as simple as possible, e.g. `Conduit_lwt.connect` should just work)
- conduit-lwt-tls and conduit-lwt-ssl (should be in separate packages, so that users could choose to install one of the other)
- conduit-async, conduit-async-lwt and conduit-asynct-tls
- conduit-mirage : just the core logic for mirage, as the rest of the functors will be applied by the tool directly.

I think it would also be nice to resurrect the mechanism in conduit 2.0 to import a local resolve.conf file.

dinosaure:
> conduit-lwt (a core subpackage, but the main unix package and the namespace should be as simple as possible, e.g. Conduit_lwt.connect should just work)

I don't understand this specific task where we have `conduit-lwt` and `conduit-lwt-unix`. Do you want a merge of them (and implicitly remove type equalities between `conduit-mirage` and `conduit-lwt-unix`?)

> conduit-lwt-tls and conduit-lwt-ssl (should be in separate packages, so that users could choose to install one of the other)

I don't have a strong opinion about that but it will comes with an explosion of OPAM packages - and I tried to avoid that as mush as I can.

tests/flow.ml
dinosaure: I try to figure out a bit about this part and I think it should be a part of `mirage-flow` instead to integrate it in `ocaml-conduit`. One crucial point is: `conduit` uses an other interface than `mirage-flow`.

A plan can be:
1) release `ocaml-conduit` with the conflict between `mirage-flow` and `Conduit.FLOW`
2) start to deprecated the `mirage-flow` interface (as mentioned into `conduit-lwt.flow`)
3) update `mirage-flow` interface and do a revdeps on it (specially, update `cohttp`)
4) implement `Loopback` into `mirage-flow` then

dinosaure: About [deprecation of `mirage-flow`](https://github.com/mirage/ocaml-conduit/pull/311/files#diff-38af78a963122b1fbe844f31d53444e9R1).

avsm:
> I don't have a strong opinion about that but it will comes with an explosion of OPAM packages - and I tried to avoid that as mush as I can.

I think the key principle behind deciding if something if a separate opam package is the dependency graph.  If they are suitably divergent, then it's worth splitting up the package. I think this is true for conduit-tls and conduit-ssl.

samoht:
Yes I would like to merge `conduit-lwt` and `conduit-lwt-unix` into `conduit-lwt`. The rational is the following: 99% of lwt users will be on unix so we should make the API as simple as possible for them:  just install `conduit-lwt` and write `Conduit_lwt.resolve ...`. The `-unix` suffix is just noise.

dinosaure:
> Yes I would like to merge conduit-lwt and conduit-lwt-unix into conduit-lwt. The rational is the following: 99% of lwt users will be on unix so we should make the API as simple as possible for them: just install conduit-lwt and write Conduit_lwt.resolve .... The -unix suffix is just noise in that case.

Ok I see your point, I will do that.

mirage/conduit_mirage_tls.ml
```diff
@@ -0,0 +1 @@
+include Conduit_tls.Make (Conduit_mirage.Mirage_scheduler) (Conduit_mirage)
```
samoht, line 1: We don't want this here as it brings a dependency to `tls` in `conduit-mirage` -- `conduit-mirage` should just define the core functors, that's the role of the mirage tool to instantiate the functors.

dinosaure:
I optimized the projection of implementations from a given abstract `flow` to be sure that `Conduit.send`/`Conduit.recv` have a small _overhead_ and I added some comments to clarify cryptic code. I did a little tool which shows to us the _overhead_ and I should implement a proper benchmark as `eqaf` - but it requires more works.

Even if we denote an overhead due to the use of an internal `hashtbl` to extract the implementation associated to the given abstracted `flow`, such design can assert a conservative approach about Conduit:
> For one given flow, it exists one and only one implementation

A more general approach exists about Conduit where we can overuse GADT and extensible variants (like [Generic Programming in OCAML](https://hal.inria.fr/hal-01664286/document)). However, it will be hard to track which implementation we use - where it would be possible to _tweak_ the associated implementation of the `flow` at any layer (the only constraint will be a type constraint between the flow and the implementation).

I started to rewrite Conduit with a conservative view where it should be easy for any to determine which value __and__ which implementation we use as Conduit determines which implementation it returns for a given flow. More technically it's the totality assumption given by `lib/e0.ml`.

Of course, this assumption has a cost, the "projection" of the module implementation from a given flow is a `hashtbl`'s lookup. However, the current API provided in this PR gives several way to extract the implementation:
- `Conduit.send`/`Conduit.recv` which do the extract systematically
- `Conduit.flow` which gives you the module implementation

We probably should write something on the documentation and, at least, write a document about how to use Conduit.

dinosaure:
> We probably should write something on the documentation and, at least, write a document about how to use Conduit.

So, on my bare-metal server, a little benchmark shows an _overhead_ of ~45ns when we use `Conduit.recv`/`Conduit.send` which seems fair and good enough. The benchmark is integrated into the distribution and it requires `clock_gettime`.

dinosaure:
So I think we reach the end of the PR where I fallback all patches and requirements listed before and I just pushed an update to `cohttp` with the new design of the API. Benchmarks was done too.

@avsm, @mirage/core should we agree to merge it and cut a release?

samoht:
I've pushed a few more cleanups commits on your branch. I still have a few question/issues but that is not blocking the merge can be added a in a separate PR:
- README.md needs to be updated (no more `CONDUIT_TLS`, new packaging, etc)
- what about `CONDUIT_DEBUG`: should we add an equivalent (/cc @avsm)
- do we want to add a default resolvers for unix? It's nice to be able to control these for mirage, but most of the time on unix it would be nice if conduit would "just work" with minimum code/configuration from the user perspective.

dinosaure:
> do we want to add a default resolvers for unix? It's nice to be able to control these for mirage, but most of the time on unix it would be nice if conduit would "just work" with minimum code/configuration from the user perspective.

The problem is: the `resolv : [ host ] Domain_name.t -> 'edn` function for `unix` (so, `'edn = Unix.sockaddr`) must contains the `port` information and a protocol which can create a resource (eg. a `socket`) must exist such as:
```ocaml
let resolv_http v = let { Unix.h_addr_list; _ } = Unix.gethostbyname v in Some (Unix.INET_ADDR (h_addr_list.(0), 80))
let resolvers = Conduit.add resolv_http my_protocol_with_sockaddr Conduit.empty
```

In the old version of `conduit`, the port was given by the service description (such as `/etc/services`) and the `'edn` value is well-know by `conduit` (but it is not extensible). With this new version, it will be hard to:
- define which port we use for which protocol
- a set of non-abstracted values which can be created and used by any protocols

So I'm not against the idea but it seems hard to give a good way to do that.

samoht:
I'm fine if we just define this default resolver for unix ; e.g. reading `/etc/services` and generating socket endpoints. We could even make `conduit-lwt-ssl` and `conduit-lwt-tls` register hooks in `conduit-lwt` for tls connection (and re-use `CONDUIT_TLS` to choose the correct one if both libraries are linked)

dinosaure:
Ok let's do that in another PR :+1:.

hannesm:
> the port was given by the service description (such as `/etc/services`)

FWIW, there's a function doing this http://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALgetservbyname -- but maybe I don't understand why this won't work / is not extensible.

dinosaure:
> FWIW, there's a function doing this http://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALgetservbyname -- but maybe I don't understand why this won't work / is not extensible.

We can imagine a use of it to _fill_ the local `resolvers` map with `resolve : [ `host ] Domain_name.t -> Unix.sockaddr`. However:
```
ssh    22/tcp
```
does not mean that the produced `resolvers` will be able to start an SSH connection. It still start only a TCP (or UDP) connection. 

> but maybe I don't understand why this won't work / is not extensible.

In the previous version of `conduit`, the ability to start a TLS connection on a specific port from an `Uri.t` is due to:
1) `conduit` __knows__ about `ocaml-tls`:
  into details, this is much more complex where `conduit` requires `ocaml-tls` _xor_ `openssl` and provides a common interface
  between these implementations to be able to use one - however, in this version, I deleted that because we have no clue about 
  which implementation we used.
2) an arbitrary constructor exists to be able to start a TLS connection
3) `conduit` implements how to create needed values to start a TLS connection

The non-extensibility comes from:
1) the arbitrary constructor (if we want to add SSH, we must extend the internal ADT of `conduit` - so make a new release)
2) the way to create needed values to start a connection
  for `ocaml-tls`, it's a `Tls.Config.client` but for SSH it's another kind of values. Finally, the old version of `conduit` must, in any way, be able to create these required values from a well-know value (in your case, the `Uri.t`).

The new version of `conduit` comes with multiple degree of liberties:
1) The internal ADT is extensible
2) The way to construct required values from a well-know value (in your case, the `[ host ] Domain_name.t`) must be define by the user

> the "old" API was basically "Uri.t -> socket", changing this to "Domain_name.t -> socket" won't work (since the protocol is then missing, which encodes a human-memoizable name for the port)

I would like to not repeat the same error. Try to fill the local `resolvers` map (which contains all of our functions which are able to create needed values from a `[ host ] Domain_name.t`) from an `/etc/services` fallback us to the third point "How to create needed values to start a TLS/'protocol connection".

At this question, several persons said that the user should control this aspect of `conduit` - and should define how to make a `Tls.Config.client`. But `/etc/services` and/or the core of `conduit` does not tell to us:
1) which implementation of TLS we want (and if we have one available)
2) and which values are needed to start a TLS connection

From my perspective, `/etc/service` can lead a part of values required to start a TLS connection but the composition from what `/etc/service` gives to us and what the user wants (a TLS connection, a SSH connection) should be done by him (and only him).
So `/etc/service` can lead the way to create an:
- `Lwt_unix.sockaddr` for `conduit-lwt`
- `Async.Socket.Address` for `conduit-async`
- `Ipaddr.V4.t * int` for `conduit-mirage`

(Note that I denotes only packages which are a well-know TCP/UDP implementation). But it should not give to user, for free, the local `resolvers` map.

dinosaure:
Any chance to have a review and/or a merge this week @avsm?

dinosaure:
So I currently deleted the CI support of Mac OSX and Windows where:
1) Installation of `libssl` on Mac OSX is a bit weird and `opam depext libssl` does not work
2) `core` does not work on Windows

dinosaure:
A high usage of `ocaml-conduit` (with unikernels) show to us (specially @hannesm and me) that `[ host ] Domain_name.t` as the concrete value to _resolve_ an endpoint is probably not the best choice when you should not require a /domain-name resolver/ at this point - not that it's not a real domain-name resolver but a simple function which should resolve a domain-name.

In the case of where we need to be connected to a peer which does not have a proper domain-name but only an IP address, it's difficult to tell `conduit` to try a connection to `10.0.0.1` for example. A hack is to decide **arbitrary** about a fake domain-name which is translated then to our IP address such as:

```ocaml
let resolvers =
  Conduit.empty
  |> Conduit.add TCP.protocol @@ fun domain_name -> match Domain_name.to_string domain_name with
  | "localhost" -> return (Some Ipaddr.V4.localhost)
  | "any" -> return (Some Ipaddr.V4.any)
  | _ -> return None
```

A solution is to replace `[ host ] Domain_name.t` by a type such as:
```ocaml
type endpoint =
  | Dns of [ `host ] Domain_name.t
  | Ip of Ipaddr.t
```

And provide finally:
```ocaml
val resolve : endpoint -> resolvers -> (flow, [> error ]) result io
```

It does not change too much the core but it's an other indirection about what the user need to give to us to start a connection (note that `Uri.t` is able to have an IP or a simple _hostname_). So, from my perspective, I agree with such change but I'm more disposal to delay this update after a merge of this (huge) PR. (@mirage/core?)

dinosaure:
So I think it's time to merge this PR, waiting the CI (which should be green). We reach the point where I can not spend much more time to maintain all PRs according this new version and @mirage/core agreed with this new version. The next real plan is to propose an other PR which will implement an `endpoint` (with domain-name and IP as explained [here](https://github.com/mirage/ocaml-conduit/pull/311#issuecomment-694970864))), fallback this change into PRs (such as cohttp and ocaml-git) and cut a release!

The merge will invalidate #319 (which requires an new version of `lwt_ssl`), #316 & #312 (move to OCaml CI), #296 (which is done by mirage/awa-ssh#19), #139 and #133 which are pretty old (4 years ago). Issues will be close step by step with an explanation about what it should be a fix (according to [this comment](https://github.com/mirage/ocaml-conduit/pull/311#issuecomment-629342906)).

avsm (approved):
Lgtm. In terms of next steps, lets:

- merge to master (I will do so now)
- add the endpoint/ip functionality in a separate PR as you suggest
- prepare a migration guide from conduit 2, which @hannesm had some ideas for as he has ported his live unikernels
- post this on discuss.ocaml.org and add to the CHANGES file to this repo
- merge the various cohttp changes (cc @mseri) into cohttp master
- then cut releases into opam of conduit and cohttp at the same time

mseri:
Awesome job!

async/conduit_async_ssl.mli
```diff
@@ -48,11 +48,11 @@ val protocol_with_ssl :
   (context * 'edn, 'flow with_ssl) Client.protocol
 
 val service_with_ssl :
-  ('cfg, 't * 'flow) Service.service ->
+  ('cfg, 't, 'flow) Service.service ->
   reader:('flow -> Reader.t) ->
   writer:('flow -> Writer.t) ->
   ('edn, 'flow with_ssl) Client.protocol ->
```
samoht: is that protocol still needed here?

Issue #310, opened 2020-03-19
Conduit.3.0.0


dinosaure:
Hi all,

I think it's time to start to discuss about a possible migration of `conduit` to the code-base of [`tuyau`](https://github.com/dinosaure/tuyau.git). More generally, one year ago we decided to find a proper way to replace `conduit` by something more powerful.

`conduit` can be hard to understand and it's why I wrote an article to explain more precisely what I had in my mind: [Tuyau, the next conduit](https://blog.x25519.net/2020-02-27--tuyau.html).

Some tasks come to my mind:

- be able to properly initialize a TLS connection (certificates, etc.)
- a better interface about `Service`
- be able to extend outside the library protocols available

I think we solved all of that into a (not so) new interface. The key of this deep update is the use of [extensible variant](https://caml.inria.fr/pub/docs/manual-ocaml/extensiblevariants.html) and GADTs.

So I think it's time to talk more technically about a possible migration, details about implementation, some missing points about comprehension (which is possible even if I wrote an article about that) and finally a plan over all MirageOS projects to migrate to the new interface.

I really would like to aggregate opinions of all people where `conduit` is everywhere and I would not like to spoil anybody about a possible migration.

The best of what you should do is to read my article (which wants to give you a global view of the project) and examples available on my repository to really understand what is going on. For an other feedback, my current blog and https://paste.x25519.net/ uses both `tuyau` instead to use `conduit` as a way to initialize a server (I still continue to use `conduit` for `irmin` however as a client). I did not get crashes at this layer from what I know.

I'm really disposal to talk about some points:
- the idea to use `[ host ] Domain_name.t` instead `Uri.t`
- enforce at the first layer type of `input` and `output`
- keep the ability to switch between `Lwt` or `Async` (or `Unix` as my example)
- the idea to use a `Map.t` (no side-effect) instead an `Hashtbl.t` as the global resolver
- the new interface of `FLOW` (eg. `val read : flow -> input -> ([ Eof | Input of int ], error) result`)
- the new interface of `SERVICE` (eg. `listen` & `accept`)
  about that, you mostly be interested by `tuyau_mirage_tcp` which re-implement `accept` to
  be close to `Unix.accept`
- the way to compose a `FLOW` with TLS (eg. `tls/tuyau_tls.ml` is the main implementation for any)
- the global design which expects that the user must fill explicitly the `Map.t` to be able to resolve instead that a filling my application of _functor_ (as `conduit` does with `functoria`)

[...]

PR #314, opened 2020-05-12
update for dns-client 4.5.0 which introduces timeouts for DNS requests

hannesm:
I'm not sure how far #311 is in respect to timeline of being merged (//cc @avsm @dinosaure). if it'll be done this week, please disregard this PR. For MirageOS use, this also requires changes to the mirage tool (since the functor arguments are hardcoded there), which I'll PR soon.

### DIFF
--- a/conduit-mirage.opam
+++ b/conduit-mirage.opam
@@ -16,7 +16,7 @@ depends: [
   "mirage-flow" {>= "2.0.0"}
   "mirage-flow-combinators" {>= "2.0.0"}
   "mirage-random" {>= "2.0.0"}
-  "dns-client" {>= "4.1.0"}
+  "dns-client" {>= "4.5.0"}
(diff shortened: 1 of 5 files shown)


dinosaure:
#310 should be merged this week, I'm waiting some reviews. About your proposal, `conduit.3.0.0` will not come with a _default_ DNS resolver (but it can provides some helpers such as [`resolv_conf`](https://github.com/dinosaure/ocaml-conduit/blob/3.0.0/lwt-unix/conduit_lwt_unix_tcp.mli#L77) for `conduit-lwt-unix`).

About `mirage`, we can choose to add such helpers into `conduit-mirage` with your DNS library or let the user to define it by himself.

hannesm:
thanks, so let's merge this and release conduit-mirage only -- and get conduit 3 released soon after :)

PR #316, opened 2020-05-26
Update test.yml

smorimoto:


dinosaure:
Thanks, `master` (with #311) has this change.

PR #319, opened 2020-06-21
use close-notify for unidirectional shutdown

madroach:
ocaml-ssl's shutdown does a full bi-directional shutdown.

one-way shutdown is needed to send `EOF`.

See also https://github.com/ocsigen/lwt_ssl/pull/2 and https://github.com/savonet/ocaml-ssl/pull/63


dinosaure:
Due to #311, this PR can not be merge as is. However, the fix still is relevant (but it needs a release of `lwt_ssl`/`ocaml-ssl`). I make an issue to keep this change in our TODO but I will close the PR. Thanks!

samoht:
Re-opening as we revert the conduit3 release.

PR #322, opened 2020-09-29
Add endpoint and replace it by [ `host ] Domain_name.t

dinosaure:


src/core/endpoint.ml
```diff
@@ -0,0 +1,39 @@
+type t = Domain of [ `host ] Domain_name.t | IP of Ipaddr.t
```
avsm, line 1: How about expanding this to be `IPv4 of Ipaddr.V4.t | IPv6 of Ipaddr.V6.t` to save some indirection? There's unlikely to be any more IP variants in the next decade :-)

hannesm: what would the advantage be? I'm in favour of `IP of Ipaddr.t`.

avsm: just a minor improvement in that it saves a boxed value. Probably not significant, except that a resolver can come back with multiple IP addresses for an endpoint.  It's a potential problem in the current `resolver` interface that only one `Endpoint.t` can come back. What should do the selection between an `A` and `AAAA` record for the same DNS entry?

hannesm: ic. I have not in-depth looked into the current resolver interface, but how to deal with multiple addresses and address families is nicely documented in "happy eyeballs v2" https://tools.ietf.org/html/rfc8305 -- I'd suggest to keep that out of this PR though (but eventually discuss how to deal with it before the 3.0 release?)

dinosaure: I think the problem about many endpoints is a good question but, as @hannesm said, we should make a new issue/PR about that.

avsm: that's fine by me - would you be able to create that followup issue @dinosaure? I've not read rfc8305 in detail yet but hope to do so early next week

src/core/endpoint.ml
```diff
@@ -0,0 +1,39 @@
...
+let pp ppf = function
+  | Domain domain_name -> Domain_name.pp ppf domain_name
+  | IP ip -> Ipaddr.pp ppf ip
+
+let error_msg str = Error (`Msg str)
+
+let error_msgf fmt = Format.kasprintf error_msg fmt
+
+let of_string str =
+  let ( >>= ) = Result.bind in
```
avsm, line 12: This bumps up the minimal OCaml requirement up to >= 4.08 by the looks of it

avsm (changes requested):

src/core/endpoint.ml
dinosaure: yes, I will start to clean issues today :+1:

src/core/endpoint.ml
```diff
@@ -9,7 +9,7 @@ let error_msg str = Error (`Msg str)
 let error_msgf fmt = Format.kasprintf error_msg fmt
 
 let of_string str =
-  let ( >>= ) = Result.bind in
+  let ( >>= ) = Stdlib.Result.bind in
```
avsm, line 12: I think just bump up the minimum OCaml version. This still fails on 4.07

dinosaure: Hmmhmm, it's a bit of shame to bump the version because we don't have `Result.bind` ... Re-write `bind` is fine for me to support more OCaml versions.

hannesm:
thanks for clarifying, @dinosaure. taking a step back, my usage is as follows:
- I use git (memory store) + irmin-kv-rw to get a in-memory clone of a remote git repository (where I'm the single writer)
- here the "remote" is passed as a (git)-URL to the application (i.e. ssh://1.2.3.4/my-repo.git - plus mutual authentication data)

now, the git repository contains authoritative information from a name server (-> a zone file, for e.g. nqsb.io), and the git server hostname is below that domain (git.nqsb.io). to avoid cycles in booting (or relying on cache mechanisms), I need to be able to specify the IP address in the git url, and not use DNS to try to resolve the hostname to the IP address. this PR may solve this issue together with some adaptions in ocaml-git. hardcoding the IP address in the unikernel itself is not an option (hard to distribute the binary then), another option would be to pass the IP address and repository name as separate boot paramerters, but that is pretty inconvenient (but would allow me to build a custom resolver and use a hostname) -- I could even parse the uri and build the resolver manually in the unikernel -- so eventually this PR is not needed (for me).

the source of the above described use case is at https://github.com/roburio/dns-primary-git/tree/ssh -- using the old conduit.

hannesm:
what I mean in my previous comment is implemented at: https://github.com/roburio/unipi/commit/2f2410bdf70e7c4f177308a4d15dec0c1bc4343f -- if the host in the uri is an ip address, use it without using any dns resolution. convenient for the unikernel user, a bit hackish (replacing the host with "reserved".

dinosaure:
So I think, from my point of view, this PR is ready to merge and even if several issues still exist about the _resolution process_, it paves the way for a better integration with MirageOS.

PR #324, opened 2020-10-01
conduit-mirage: support dual IPv4 and IPv6 stack

hannesm:
as outlined in the diff, this needs a pin of the dns-client


samoht:

(or maybe re-open something against `master`?)

PR #327, opened 2020-10-09
Add a simple HOW-TO which explain how to change a protocol with Conduit

dinosaure:


hannesm:
I recently came across http://reo.project.cwi.nl/v2/reowolf/api.html and was wondering whether you have some remarks / what is different & similar to conduit 3.0.0. may serve as inspiration or put into a "related work" section

dinosaure:
> I recently came across http://reo.project.cwi.nl/v2/reowolf/api.html and was wondering whether you have some remarks / what is different & similar to conduit 3.0.0. may serve as inspiration or put into a "related work" section

I'm not sure, I can see some common idea - like "BSD socket is not systematically available so we must abstract it" - but it seems that they go to another direction (with different kinds of protocols).

dinosaure:
The HOW-TO still is relevant according to last PR (to fix `cohttp`) - I tested the code and it's working. So the PR is still relevant.

dinosaure:
So, I recheck the HOW-TO and it seems clear about the way to use `Conduit` with examples and a concrete goal: a simple ping-pong server where we want to upgrade to TLS in the most easy way. I added a paragraph about MirageOS to talk a bit about _functors_ and let the `readme.mld` as a more complete (but less easier) explanation of `Conduit`. I think it's ready to merge.

PR #328, opened 2020-10-12
Fix use of lwt_io + conduit-tls

dinosaure:
According to the documentation of `conduit-tls`, the implementation with `ocaml-tls` is not thread-safe, something like `both (send flow) (recv flow)` is unsafe. It's mostly about the wrapping of a `Conduit.flow` with `Lwt_io` - because the `flow` can be a TLS flow, we must protect `recv` and `send` with a mutex anyway.

The question behind such patch is: should we protect `conduit-tls` with a mutex - and by this way, we are not mandatory to protect any `flow` with a mutex or should we keep this design? To this existential question, we should make an issue a talk about this issue later - note that this is not a part of the core of `conduit`.

The second patch is about the _first-read_ behavior of `Lwt_io`. I'm not sure how we can properly schedule that but it seems that the `Lwt_io` wants to read first when we start a TLS connection - I suspect that it schedule the action from the underlying state of the socket and, because we receive a part of the handshake, it considers `Conduit.recv` as the first operation. However, for the HTTP/1.1 point-of-view, this is not true when the server expect to receive something - so client and server expect to receive something at the end.

`conduit-tls` is able to produce an opportunity to schedule `Conduit.send` even if we called `Conduit.recv`, it returns `Input 0` (which means that we received nothing, __but__ the connection still is established). On the other side, `recv` given to `Lwt_io` can _yield_ to give the opportunity to the application to scheduler `Conduit.send`/`Lwt_io.write`.

Finally:
1) the question about `conduit-tls` is a bit hard but an issue will be done which describe precisely the situation
2) the problem with `lwt_io` is an inheritance of 1) and it seems that `tls.lwt` fix it in its own way

### DIFF
--- a/src/lwt/conduit_lwt.ml
+++ b/src/lwt/conduit_lwt.ml
@@ -113,8 +113,28 @@ module TCP = struct
       socket : Lwt_unix.file_descr;
       sockaddr : Lwt_unix.sockaddr;
       linger : Bytes.t;
+      recv_first : bool;
       mutable closed : bool;
     }
+
+    (* XXX(dinosaure): [recv_first] is here to fit into [Lwt_io], from what we know,
+     * a tuple of [Lwt_io] [in_channel/out_channel] tries to receive first. However,
+     * such behavior is problematic for HTTP:
+     * - as a HTTP client, we should send first
+     * - as a HTTP server, we should recv first
+     * - with TLS layer [conduit-tls], both work - where
+     *   the handshake can be done by send or recv
+     *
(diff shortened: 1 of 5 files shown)


dinosaure:
Ok, it seems that the first-recv behavior of `Lwt_io` is true in any case (TLS or not), the second commit gives the opportunity to reschedule `Conduit.send` with an `Input 0`.

dinosaure:
The real conclusion of this patch is: don't use `Lwt_io` when you need a full control of how to schedule `recv`/`send`...

PR #331, opened 2020-10-15
Update the README.md

dinosaure:


dinosaure:
After some feedbacks (thx @ulugbekna), it seems that one of my document is better than the current `readme.mld` (which is too big and less comprehensive). The mix between the HOW-TO (which describes precisely the API) and the new document (which describes the goal of Conduit) is a good mix to show Conduit.

PR #332, opened 2020-10-15
howto: no need to manually initialize the rng (conduit_lwt_tls does this for us)

hannesm:
the initialization via Mirage_crypto_rng_unix.initialize is meant for non-lwt programs, since there is only an initial seeding of the RNG, and no periodic reseeding.

Since this example / howto is well-advertised, let's remove the manual initialization -- the conduit-lwt-tls package already takes care of it (using Mirage_crypto_rng_lwt.initialize -- which periodically collects entropy from getrandom/getentropy and rdrand/rdseed).

### DIFF
--- a/src/core/howto.mld
+++ b/src/core/howto.mld
@@ -176,8 +176,6 @@ we will use them to upgrade our protocol (when the scheme is ["pgs"]):
 
 {v server.ml v}
 {[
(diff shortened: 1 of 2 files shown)


hannesm:
thanks for merging, would imho be good to update the gh-pages branch as well :) -- so that the linked documentation / howto.html is up-to-date.

PR #335, opened 2020-10-28
Allow json benchmark output 

gs0510:
& add bench target to Makefile

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,8 +24,9 @@ jobs:
         opam pin add -n conduit-async-ssl.dev .
         opam pin add -n conduit-lwt-tls.dev .
         opam pin add -n conduit-lwt-ssl.dev .
+        opam pin add -n conduit-bench.dev .
         opam depext  -y conduit conduit-tls conduit-lwt conduit-async conduit-mirage
-        opam install --deps-only -t conduit conduit-tls conduit-lwt conduit-lwt-tls conduit-lwt-ssl conduit-async conduit-async-tls conduit-async-ssl conduit-mirage
+        opam install --deps-only -t conduit conduit-tls conduit-lwt conduit-lwt-tls conduit-lwt-ssl conduit-async conduit-async-tls conduit-async-ssl conduit-mirage conduit-bench
     - name: Build
       run: opam exec -- dune build
     - name: Test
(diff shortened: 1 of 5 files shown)


dinosaure (changes requested):
I'm sorry but `conduit` is deep in the stack and we should not introduce a new dependency on it 😕 . Can you produce a `json` output without PPX and Yojson?

avsm:
This is only for benchmarking though: can't it be a depopt?

dinosaure:
Hmmhmm I'm not sure in which context the benchmark is running (it can run as a _test_ and, in this case we can tag this dependency with `{with-test}`). Otherwise, a `depopts` with `(optional)` is fine - but, in that case  we should extend the CI to ensure that the benchmark still builds.

hannesm:
I don't know much about your benchmarking setup, but please note that we have https://github.com/mirage/metrics specifically for such a use case "separate the reporting from the collection". With metrics, the "core" library is very slim (and should not be an issue to integrate into your library - exposing one or many metrics sources), while the reporter can be installed by the application - which can be a simple text outputter, or a json or sexp or whatever outputter. IMHO the metrics library (similarly to logs) separates the collection from the reporting, and makes a lot of sense in respect to dependency cone.

samoht:
What we've done in [other packages](https://github.com/mirage/irmin/blob/master/irmin-bench.opam) is to have an explicit `*-benchmark` opam package to hold these kinds of dependencies. Maybe that would help to do the same thing here?

gs0510:
> Hmmhmm I'm not sure in which context the benchmark is running (it can run as a _test_ and, in this case we can tag this dependency with `{with-test}`). Otherwise, a `depopts` with `(optional)` is fine - but, in that case we should extend the CI to ensure that the benchmark still builds.

I would be running the `bench` target from the Makefile, but would only want to run the benchmarks and not other tests. Is there a solution out of the ones listed here that is more preferred than the others?

samoht:
I think we should avoid using depopts dependencies as much as possible -- it brings non-determinism and a lot of recompilation to our users.

I think we shoud go with creating an additional `conduit-bench` target ; Looping in @CraigFe who has already had to solve these issues for irmin and index: we should solve that issue in a consistent way to ease the job of the continuous benchmark infrastructure that @gs0510 is working on.

gs0510:
@samoht I talked to @dinosaure offline, and he recommends creating a `conduit-bench.opam`, so I'll go ahead with that since it would also make things more uniform.

dinosaure (approved):

dinosaure:
Thanks!

PR #337, opened 2020-11-18
Conduit_{lwt,async}.serve does not need the protocol anymore

dinosaure:
After a discussion about `cohttp`, it seems that the rational behind this signature:
```ocaml
val serve : 'cfg
  -> protocol:(_, 'flow) Conduit.protocol
  -> service:('cfg, 'service, 'flow) Conduit.Service.service
  -> handler:('flow -> unit io)
  -> unit io
```

is not so clear and can lead the user to some compilation errors (when `protocol` and `service` don't have the same `'flow` type) or runtime errors[1]. This PR is a revert of [this specific commit](https://github.com/mirage/ocaml-conduit/pull/311/commits/c84a6b45ecbe0b6846dd48d9619ab2c0df3220a2) which comes from a nightly agreement with @samoht.  But it seems that the final result is not good according to this comment: https://github.com/mirage/ocaml-cohttp/pull/732#discussion_r518744814

This PR provides an new interface (so, it breaks the current API):
```ocaml
val serve : 'cfg
  -> service:('cfg, 'service, _) Conduit.Service.service
  -> handler:(Conduit.flow -> unit io)
  -> unit io
```

Note that if the user wants to directly get the concrete value `'flow`, he is able to do that with `Conduit.Service.impl service` and the real implementation of the service (with concrete types). With the new interface of `serve`, as a Conduit's protocol, the user is still able to _destruct_ the value to its concrete type.

For the `cohttp`'s point-of-view, it will __just__ delete the `protocol` argument when we want to launch a CoHTTP server.

About release, Conduit 3.0.0 is currently not used by any released packages, so:
- we can lie and redo the release of Conduit 3.0.0
- we can do a Conduit 3.1.0 and explain the rational under this API update

cc @mirage/core @avsm @samoht 

[1]: a runtime error can exists when the user wants to serve with a service and a protocol and both don't come from the same place (and don't share the same logic, axiom, etc on their implementations but they share the same type).

### DIFF
--- a/src/async-ssl/conduit_async_ssl.ml
+++ b/src/async-ssl/conduit_async_ssl.ml
@@ -284,7 +284,7 @@ let service_with_ssl :
     writer:(flow -> Writer.t) ->
     (edn, flow with_ssl) Conduit_async.protocol ->
     (context * cfg, context * t, flow with_ssl) Conduit_async.Service.service =
- fun service ~reader ~writer _ ->
+ fun service ~reader ~writer protocol ->
   let module S = (val Conduit_async.Service.impl service) in
   let module Service = struct
     include S
@@ -294,7 +294,7 @@ let service_with_ssl :
     let writer = writer
   end in
   let module M = Make (Service) in
-  Conduit_async.Service.register ~service:(module M)
+  Conduit_async.Service.register ~service:(module M) ~protocol
 
 module TCP = struct
   open Conduit_async.TCP
--- a/src/async/conduit_async.ml
+++ b/src/async/conduit_async.ml
@@ -16,10 +16,10 @@ let ( >>? ) x f = Async.Deferred.Result.bind x ~f
 type ('a, 'b, 'c) service = ('a, 'b, 'c) Service.service
 
 let serve :
-    type cfg t flow.
+    type cfg t v.
     ?timeout:int ->
     handler:(flow -> unit Async.Deferred.t) ->
-    service:(cfg, t, flow) service ->
+    service:(cfg, t, v) service ->
     cfg ->
     unit Async.Condition.t * (unit -> unit Async.Deferred.t) =
  fun ?timeout ~handler ~service cfg ->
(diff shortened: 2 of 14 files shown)


samoht:
I've been reading the wrong PR 🤦 

To ease review, can you try to keep the same variable names when possible? Thanks!

samoht:
I understand that this PR moves the burden of providing a protocol to the registration site instead of to the use site. I think it is already a major improvement. What is not clear to me is why a service registration needs a protocol at all: services do not need to call `connect` so they should not need it.

dinosaure:
The protocol witness given at the registration service permits us to _pack_ the given created and initialized flow to a `Conduit.flow` (so, yes, `Protocol.connect` is not used).

In others words, to be able to _pack_ a `'flow` to a `Conduit.flow`, we need the protocol witness (which contains the _constructor_).

About "why we need to _pack_ the `'flow`?", it's about abstraction. If a server logic (such as `ping-pong` or CoHTTP) wants to be abstract over the kind of the protocol (TCP or TLS or whatever), it must use `Conduit.flow`.

This PR highlights only one and unique point: who take care about the _pack_ operation? The user (the current state of `conduit.3.0.0`? Conduit (the current state of this PR)?

But, in anyway, this process needs the protocol witness. At this stage  we can not do lot of magics. However:
1) the registration of service is generally coupled to the registration of the protocol (as we have a client, we should have a service too)
2) At the composition level (eg. TLS), the registration needs the protocol witness (regardless this PR) - to fix the type of the `'flow`

dinosaure:
> To ease review, can you try to keep the same variable names when possible? Thanks!

Unfortunately, this is not possible when I mostly use the `flow` type variable and I would like to replace it by `Conduit.flow` type - so OCaml complain about `flow` and `'flow`, I decided to use `'v` and `v` to fix this kind of error.

samoht:
What I find confusing is that we are interchanging protocols and flows for the server implementation; I understand that the server-side implementation needs a notion of `flow` when defining a service. For me it would make sense to either pass a `flow` argument or to somehow hide that flow into the service value. `serve` doesn't care about `connect` so we should not use protocols there.

dinosaure:
> For me it would make sense to either pass a flow argument or to somehow hide that flow into the service value. 

I missed may be something but how we could do that or how it's different from what this PR provide? `serve` (or the `service` - depends on which layer we want to integrate that) needs to know the OCaml constructor to be able to make `Conduit.flow`, and the only value which can give to us this constructor is the _protocol witness_.

samoht:
Merging this -- will integrate changes from #338 as separate PRs.

PR #338, opened 2020-11-26
Remove protocol argument to Service.register

samoht:
The core of the change is 3286638: this commit keeps a witness for flows, protocols and services (and a link between them). It's a bit larger than I have expected, but it's also cleaning up the Service API a bit by making it an extension of a flow (and hence I had to rename `Service.close` into `Service.stop` to avoid name clashes).

The rest of the changes are just cosmetic. The main one being the removal of labelled arguments (as we just have one argument now).


dinosaure:
I think I understand your PR - not so easy project - however, I have some concerns about `repr` between a protocol and a service. I mean, for any `Flw.prj`, you create a new constructor and I think that the constructor exposed (like in `Conduit_lwt.TCP.T`) is not equivalent to the constructor used to make the _packed_ flow (in `accept`).

In others words:
```ocmal
let handler = function
  | Conduit.Value (Conduit_lwt.T flow) -> ... (* XXX(dinosaure): never appear *)
  | flow -> ..

let run = Conduit_lwt.serve ~handler ~service:Conduit_lwt.TCP.service cfg
```

samoht:
Hum do we have a way to test if this is broken or not? It's indeed a tricky codebase so I have no idea if I broke this or not :-)

dinosaure:
I added a test on `test/flow.ml` which highlights the case. Usually, we expose the constructor from the protocol. However, as I said, registration of a service creates a new constructor for the flow. So, we don't have a way (currently) to _destruct_ the flow produced by `accept` to what it really is.

The test should be easily to patch-back to `trunk` - but it will be slightly different. I can make a PR about that if you want - to show the real diff.

samoht:
That seems to work if I add a `Service.repr` function. Need to think a bit more if you think `Conduit.repr` is also supposed to work on protocol witness (I'm not sure they should).

dinosaure:
> That seems to work if I add a Service.repr function. Need to think a bit more if you think Conduit.repr is also supposed to work on protocol witness (I'm not sure they should).

Depending on the resulting API. I mean, with your last patch, you assume: two separate constructor for a _flow_ depending on where it comes from. `protocol_repr` is when the user starts a _flow_ with `Conduit.resolve` and `service_repr` is when the user starts a _flow_ with `Conduit.Service.accept`.

I'm not sure that this diff is clear (even if a rational exists behind) and it can lead errors - such as: my value is not correctly destructed.

An other problem is the impossibility to expose both constructors into the same module (otherwise, you will get `Multiple definition of the extension constructor name T`) - which is already the case without this PR but I did not think that we would have several `repr` into one module.

A possible solution (but not tested and I don't know if it's really applicable) is to produce 2 kind of constructor such as:
`Flw.inj0` and `Flw.inj1`. One will create a constructor `P`, and the other will create a constructor `S`. But I'm not sure about such solution - and it can lead some others problems on the API.

samoht:
> Depending on the resulting API. I mean, with your last patch, you assume: two separate constructor for a flow depending on where it comes from. protocol_repr is when the user starts a flow with Conduit.resolve and service_repr is when the user starts a flow with Conduit.Service.accept.

That seems like a good distinction (e.g. you have client-side and server-side flows). But I agree the failure mode is not ideal as even if everything type-check you will have a runtime error in some cases. I was somehow expected that `E0.ml` would help by not duplicating entries with the same key, but maybe there is no good way to solve this. Need to think a bit more.

samoht:
I've pushed a new commit ; this one create a witness for flows that you have to pass around when registering protocols and services. The API is a bit more verbose but more "correct". The tests now pass.

There is still something a bit odd: it's not fully clear which `send` and `recv` functions are used ; `Flow.send` and `Flow.recv` will use the one registered when creating the corresponding flow witness. Otherwise it will use the one in the corresponding Protocol or Service modules. I think that's fine, but that can probably lead to some weird error (I haven't managed to find such a case, but I imagine that could happen). I've tried to remove `send`, `recv` and `close` from `Protocol` and `Service` but then we need to handle incompatible error types, which is possible but even more verbose.

So I think I am reasonably ok with the current state of that PR. @dinosaure what do you think? If you are ok with the changes, I'll add a section in the changelog to explain the API changes.

dinosaure:
I think that it would be good to take more global view on what you did and what you understood of it: the service needs to know (in any circumstance) the _constructor_ of the `flow` to be able to pass an _abstract_ value to the end-user. If we assume that, the remaining question: how we can get this _constructor_ in `Conduit.accept`?

The current status of `conduit` provides an API which does not deal with that at the core level. But we need to abstract `'v`/`'flow` to a `Conduit.flow`/`Conduit.Flow.t` (regarding CoHTTP). So, `Conduit_{lwt,async}.serve` provides a certain use of `Conduit.accept` with some arguments to deal with this problematic. Currently, `Conduit_{lwt,async}.serve` needs the `protocol` witness to know which _constructor_ can be used - and it seems odd for us (or the rational behind that requires a knowledge of Conduit's internals)

So, I came with an other solution where the registration of the service requires a `protocol` witness - in sense of a service needs to know what it wants to provide (by the `accept` function) to the user (see #338) - I would like to say that it was the initial idea. 

From my perspective, your PR redo slightly the same when, instead to re-use the `protocol` witness:
1) a third sub-module is needed which is like the `Protocol` module without `connect`
2) it makes a new `flow` witness
3) which is required for the registration of a service

The main diff between this PR and #338 is that you ensure 2 different implementations (but pretty-close) where one of them does not implement the `connect` function (this one, then, must be registered and used to make a service). I mean, you finally get:

https://github.com/mirage/ocaml-conduit/blob/abc9b7df51eefd62db880ee60f676f31cfba64d6/src/core/conduit.mli#L388-L389

When I did this:

https://github.com/mirage/ocaml-conduit/blob/c72831c8462c19f1445aad31158abb3eb261cbb8/src/core/conduit.mli#L348-L351

It's pretty-much the same but the first imply 2 registrations (and 2 different _constructors_ which can not leave into the same place - as explained [here](https://github.com/mirage/ocaml-conduit/pull/338#issuecomment-734317263)).

My question is simple: do we really need to enforce the fact that a service must require a _flow_ which does not implement `connect` (including extra-processes)? In a certain point-of-view, it's more convenience to implement a protocol and re-use it to register the associated `service`. The _bad_ case (in #338) is when we want to implement a service **without** its protocol (which is unusual by definition - it's like a half-job - and usually, we start to implement the client-part).

So I understand what you did but may be it will brings more complexity comparing to #338 - I mean, by usage, we don't really need to enforce the usage of a third module just to explicit the fact that a service will not use `connect`, at least, we can just explain that in the documentation (to open the door for the _bad_ case - and I'm not really sure that it corresponds to an usual case) and be less verbose on a already-complex project.

samoht:
To clarify a bit my changes, today conduit has 3 main concepts:
- flows (something users can read and write) to. This corresponds to the `FLOW` module type.
- protocols: a flow + a connect function (with an endpoint type). This corresponds to the `PROTOCOL` module type.
- services: a flow + a init (with a configuration type) and accept function. This corresponds to the `SERVICE` module type.

There's also some magic to make flows extensible and abstract. That magic uses type witness. Currently, there are 2 kinds type witnesses:
- `type ('endpoint, 'flow) protocol` to model implementations of the `PROTOCOL` type.
- `type ('configuration, 'server_state, 'flow) service` to model implementations of the `SERVICE` type.

In conduit3, you currently need to pass a witness of a protocol and a service to start a service. This is a bit hard to document and explain, as you look at the type, you would expect to just need a `SERVICE` implementation (services do not need to call `connect`!)

- #337 is moving the protocol witness at the registration time, so that `serve` would just take a service witness.
- #338 (this PR) is creating a new `flow` witness, also to be used at registration time, for both protocols and services

I understand your point that maybe we don't need that extra complexity of adding a new type witness. On the other hand, this simplifies the destruct API (as there is no need to expose an extra `('a, 'b) value` constructor). Will need to think a bit more :-)

In any case, there are few API simplifications that I thing we should do (like the removal of labelled arguments) for conduit 3.1.

dinosaure:
I see your point about `flow`/`protocol`/`service` but, as I said, I prefer to have less complexity - even if in the _bad_ case, we need to do `let connect _ = assert false` on our protocol implementation. The path seems too long to start to have something which works (3 modules, 3 registrations) and it is already too long.

I understand that your API is more strict that the current one where we have a clear separation between what is about `recv`/`send`/`close`, `connect` and service's stuffs. But I believe on the duality between the protocol and the service (and the second should not exists without the first one - at least, this duality ensures that we think protocol implementation in both ways - I can imagine a `Protocol` with `Flow0` and `Service` with `Flow1` with your PR).

> On the other hand, this simplifies the destruct API (as there is no need to expose an extra ('a, 'b) value constructor).

I agree that is an issue :+1:.

> In any case, there are few API simplifications that I thing we should do (like the removal of labelled arguments) for conduit 3.1.

And yes, I agree that this PR comes with some others updates (which can be integrated into some others PRs :+1:).

samoht:
So I've removed the `flow` witness and I've simplified more stuff. I've tried to remove the "exotic" functions from the public API -- I think now it's a bit easier to grasp.

Will clean-up the patches and resubmit them individually over the weekend (if I find some time).

samoht:
Closing now as everything has every useful commits have been upstreamed separately.

PR #343, opened 2020-11-29
Simplify Conduit.repr

samoht:
Remove the intermediate value type which was not necessary

Extracted from #338


dinosaure:
I re-integrated the test to check type equality from #338.

PR #344, opened 2020-11-29
Remove pack

samoht:
(on top of #343)


dinosaure:
Agreed as long as `Conduit.connect` returns a `Conduit.flow`.

Issue #356, opened 2020-12-01
Conduit.S.PROTOCOL does not exists


dinosaure:
I'm not sure if it's a nice idea to use the trick about `_intf` because, currently we can introspect `conduit.mli` and see `Conduit.S.PROTOCOL` but we can not re-use it:
```sh
$ ocaml
# #show Conduit.S.PROTOCOL ;;
Unknown element.
# 
```

Any idea @samoht?

avsm:
Shouldn't that be `#show_module_type` ?

dinosaure:
> Shouldn't that be #show_module_type ?

The error is just more precise than a call with `#show`.

CraigFe:
Unfortunately, OCaml top-level and Odoc aren't quite clever enough about resolving aliases (and the `@inline` annotation applies to inclusions only, not aliases). So we have:

### Bad

```ocaml
module type S = Conduit_intf.S
(** @inline *)
```

### Good

```ocaml
module type S = sig
  include Conduit_intf.S
  (** @inline *)
end
```

I'll prepare a patch.

CraigFe:
Should be resolved by https://github.com/mirage/ocaml-conduit/pull/357.

avsm:
cc @jonludlam on this -- this seems like something to fix in odoc as well.

CraigFe:
I believe this is already fixed in Odoc (2.0+), which expands aliases to hidden modules rather than generating documentation for them. So once we're using the new model, the above examples will be equivalent to just:

```ocaml
module type S = Conduit_intf.S
```

@jonludlam may wish to correct me on this.

avsm:
So to fix this without codesmell in our interfaces, we also need to apply the odoc model resolver to the toplevel, by the sounds of it.

dinosaure:
I'm not sure that #357 is the right direction when this simple code can not compile:
```ocaml
let _ = match Conduit_lwt.Service.equal Conduit_lwt.TCP.service Conduit_lwt.TCP.service with
  | Some (Conduit_lwt.Refl, _, _) -> ()
  | _ -> ()
```

dinosaure:
#361 re-fix/re-export the `Refl` constructor to be usable outside of Conduit.

PR #357, opened 2020-12-01
core: expand aliases exposed in _intf file

CraigFe:
This ensures that the OCaml toplevel is able to introspect these signatures directly  and prevents references to module types in `Conduit_intf` from escaping to the user.

Fixes https://github.com/mirage/ocaml-conduit/issues/356.

### DIFF
--- a/src/core/conduit.mli
+++ b/src/core/conduit.mli
@@ -1,22 +1,2 @@
-module Endpoint = Endpoint
-
-type resolvers
-(** Type for resolvers map. *)
-
-val empty : resolvers
-(** [empty] is an empty {!resolvers} map. *)
-
-module type S = Conduit_intf.S
+include Conduit_intf.Conduit
 (** @inline *)
-
-module type IO = Conduit_intf.IO
-(** @inline *)
-
-module type BUFFER = Conduit_intf.BUFFER
-(** @inline *)
-
-module Make (IO : IO) (Input : BUFFER) (Output : BUFFER) :
-  S
-    with type input = Input.t
-     and type output = Output.t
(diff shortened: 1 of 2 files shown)


dinosaure:
The problem persist... (I use OCaml 4.10.0).

CraigFe:
Which definitions are you not able to access? Things seem to be working fine on my machine.


```
$ dune utop

utop # #show Conduit.IO ;;
module type IO =
  sig
    type +'a t
    val bind : 'a t -> ('a -> 'b t) -> 'b t
    val return : 'a -> 'a t
  end
```

CraigFe:
Ah, you're expecting to be able to access all modules types under `Conduit.S`? If so, we just need to re-arrange slightly in `Conduit.Conduit_intf`; exposing `Conduit_intf` directly in the public API doesn't seem like a good idea to me.

dinosaure:
Yes, internals of `S` must be accessible, if you have a proper solution, I will happy to merge 👍

CraigFe:
I've pushed another commit that exports three more module types that were not public before:

- `Conduit.FLOW`
- `Conduit.PROTOCOL`
- `Conduit.SERVICE`.

It's a little boilerplate-y, but doing it this way preserves a sensible ordering in the docs.

CraigFe:
Pushed another commit that subsumes https://github.com/mirage/ocaml-conduit/pull/361 to avoid conflicts if you're happy with this approach.

samoht:
LGTM I think it's indeed a good idea to remove the indirection to `S` and expose these top-level signatures directly.

PR #358, opened 2020-12-01
Tls thread safe

### LINKED ISSUE #336: conduit-tls must be thread-safe
The current implementation of `conduit-tls` is not thread-safe and we must apply an exclusion between `Conduit.send`/`Conduit.recv`. Such assumption appear when the _handshake_ is done while we want to `send`/`recv` something - it differs from `tls.{lwt,mirage}` which _drain_ the _handshake_ **before** any `send`/`recv`. Two solution are possible:
1) Integrate a mutex into `conduit-tls` (which is _scheduler-free_) to be thread-safe in anyway
2) rewrite `conduit-tls` so that we _drain_ the _handshake_ at the beginning and we don't need to figure safe concurrent `read`/`write` until we don't have a renegotiation.

I'm for the second proposition. The second proposition does not fix the issue about renegotiation and the end-user (as `tls.{lwt,mirage}`) must take care about exclusion between `send`/`recv` if he want to renegotiate with the server (and vice-versa).

dinosaure:
Try to fix #336


samoht:
Do you mind cleaning up the history a little? The PR is very hard to review in its current form.

dinosaure:
Yes, it's draft I'm currently on a split of this PR.

dinosaure:
See #360, #361, #362 and #363.

PR #361, opened 2020-12-02
Re-export Conduit.Refl constructor

dinosaure:
Until #357 we need to re-export correctly `Refl` because it is used by CoHTTP and required when we want to _pattern-match_ the result of:
https://github.com/mirage/ocaml-conduit/blob/7375b559cde825e66b0805129f51464174f20eb1/src/core/conduit_intf.ml#L416-L429

### DIFF
--- a/src/core/conduit.mli
+++ b/src/core/conduit.mli
@@ -1,2 +1,4 @@
+type ('a, 'b) refl = ('a, 'b) Conduit_intf.refl = Refl : ('a, 'a) refl
+
 include Conduit_intf.Conduit
 (** @inline *)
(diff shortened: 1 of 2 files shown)


CraigFe (approved):

samoht:
Do you have an example where `Service.equal` is useful? I'm wondering what is used for.

dinosaure:
You can have an example [here](https://github.com/mirage/ocaml-cohttp/blob/4b834d57ab24b94791be9b4b68e228a83139d85a/cohttp-lwt-unix-nossl/src/server.ml#L97-L100) where a argument is specific to a certain service implementation.

PR #362, opened 2020-12-02
Better lwt tcp protocol

dinosaure:
This code will break the `ping-pong` test but it is a part of several incoming PR about the TLS stack. This update is requested from @talex5 about a clear behavior of `recv` and `send`. The documentation was updated according to the new behavior and we ensure to call only one and unique time underlying `Lwt_unix.{read,write}` _syscalls_. 

This PR is less error-prone and less complex than before. It delete a weird case `Input 0` (which was needed to call `yield ()`).


dinosaure:
Ah and it deletes the use of a mutex to make a proper `Lwt_io` tuple (used then by CoHTTP).

PR #363, opened 2020-12-02
Tls thread safe

dinosaure:
This update want to assert a thread-safe implementation of the TLS layer. The implementation comes from `tls.lwt` and was copy/paste (minor mirleft/ocaml-tls#422, a functor to unlock the ability of the composition and `ocamlformat`). CI will fails and we need to merge #362.

An other aspect is the needed unencrypted way to communicate to the client from the server to _drain_ the handshake. Composition is bit more complex than before to make a TLS service from _a protocol_: the service requires the protocol with TLS (as explained in #337 and #338 to be able to wrap the flow into the right _constructor_) and it requires the protocol **without** TLS to be able to _drain_ (and use `recv`/`send`) the handshake before to return to the end-user the flow.

### DIFF
--- a/src/async-tls/conduit_async_tls.ml
+++ b/src/async-tls/conduit_async_tls.ml
@@ -1,12 +1,28 @@
 open Async
-include Conduit_tls.Make (Conduit_async.IO) (Conduit_async)
+
+module IO = struct
+  type +'a t = 'a Async.Deferred.t
+
+  let bind x f = Async.Deferred.bind x ~f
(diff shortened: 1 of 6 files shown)


dinosaure:
With this PR and #362, CoHTTP and [paf](https://github.com/dinosaure/paf-le-chien) still works with a minor update (due to @samoht's PR) and a deletion of the mutex usage due to the thread-safety :tada:.

PR #364, opened 2020-12-02
Remove unused Conduit.S.resolvers type

CraigFe:
The internal implementation ensures `Conduit.resolvers` = `Conduit.Make(...).resolvers`, but this isn't exposed to the user so I think this type is useless.

### DIFF
--- a/src/core/conduit.ml
+++ b/src/core/conduit.ml
@@ -70,6 +70,10 @@ end) : BIJECTION with type +'a s = 'a Functor.t = struct
   external prj : ('a, t) app -> 'a s = "%identity"
 end
 
+module type S = sig
+  include S with type resolvers := resolvers
+end
+
 module Make (IO : IO) (Input : BUFFER) (Output : BUFFER) :
   S
     with type input = Input.t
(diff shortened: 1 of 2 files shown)


dinosaure:
This is probably the most important type of Conduit 😄 I mean the user should be able to fill and use it:

https://github.com/mirage/ocaml-conduit/blob/8e7283bd4804d8533e395cd9c50c5e14c057a5c0/src/core/conduit_intf.ml#L319-L344

but I agree that the corresponding constraint does not appear - and it is a problem. In any application of `Make`, the resulting `resolvers` should be equal to the `type resolvers` defined outside the _functor_.

CraigFe:
OK, will take a look at this tomorrow :slightly_smiling_face: Would be nice to have a test that witnesses this lack of type equality as an issue.

CraigFe:
I've changed the approach to avoid exporting a `Conduit.S.resolvers` type instead. `Conduit.S.empty` now has type `Conduit.resolvers` and is provided as an alias of `Conduit.empty` for convenience.

dinosaure (approved):

dinosaure:
I did the REVDEPS by hands all works with CoHTTP and HTTP/af.

CraigFe:
Thanks; will do those rev-deps from now on :+1:

Issue #352, opened 2020-11-30
Use Lwt_switch instead of Lwt_condition to stop the server loop


dinosaure:
See #339, /cc @talex5:
> Hmm, there is a Conduit_lwt.serve. But it returns a Lwt_condition.t to stop it. That's racy - if you broadcast on stop before the loop calls wait then it will miss the event! And it really shouldn't be calling wait each time around the loop.
> 
> The normal Lwt way to do that would be to take Lwt_switch argument and abort the loop when the switch is turned off.
A switch can only be turned off once and so you can't miss the event. Also, the switch argument can be optional in the common case where you never want to stop.

PR #369, opened 2020-12-04
Avoid shutdown race condition on service startup

CraigFe:
Fix https://github.com/mirage/ocaml-conduit/issues/352. The `lwt` and `async` service loops previously took condition variables that were used to signal when to shutdown. This resulted in a race condition: the shutdown signal can be broadcast before the server is waiting on it.

Fixed by having the service loops take switches as configuration instead:

```ocaml
val Conduit_lwt.serve   : ?stop:Lwt_switch.t    -> ... -> cfg -> unit Lwt.t
val Conduit_async.serve : ?stop:unit Deferred.t -> ... -> cfg -> unit Deferred.t
```
The most closely corresponding definitions would probably have `Conduit_async.serve` taking an `Ivar.t` instead, but from a glance at similar service loops in the ecosystem it seems to be more common to use a `Deferred.t` instead. Happy to change that if you think it's more appropriate.


dinosaure (approved):
May be, in your case in Irmin, it can be interesting to add a new argument `launched` which signals that the server was properly launched after that? Note that it could useful for us, in our tests, when we need to `sleep` a bit with `async` before to connect clients.

CraigFe:
I spent a while thinking about ways to know that the server is ready to receive messages, and I'm not sure it can be done with `Lwt`. As you say, we would want to emit a signal after the first time that `accept` blocks, but  without knowing how `accept` is implemented we can't pull any tricks with the scheduler to get a signal out after that thread sleeps (e.g. it might contain several non-blocking portions, and we don't know how many).

I'd be very happy to be proven wrong on this though; currently this seems like a pretty annoying limitation of Lwt and of concurrency monads more generally.

dinosaure:
> I spent a while thinking about ways to know that the server is ready to receive messages, and I'm not sure it can be done with Lwt. As you say, we would want to emit a signal after the first time that accept blocks, but without knowing how accept is implemented we can't pull any tricks with the scheduler to get a signal out after that thread sleeps (e.g. it might contain several non-blocking portions, and we don't know how many).

`Service.init` should be enough. The main error is when a client wants to connect to a specific port which is not yet bound. In a certain sense, `Service.init` _allocates_ the ressource and initialise it - so, about `Conduit_lwt.TCP` for example, it allocates the _socket_ and `bind_ it. At this stage, you can launch a client cooperatively.

CraigFe:
Good point. I've done as you suggest in https://github.com/mirage/ocaml-conduit/pull/371.

samoht:

PR #371, opened 2020-12-07
Add Service.serve_when_ready

CraigFe:
... which is like `serve` but returns ``[ `Initialized of unit io ] io``, rather than a plain `unit io` in order to allow the user to perform actions after initialisation but before the service loop begins.

In particular, this is helpful for testing a client/server pair; as is done in the tests for Conduit itself.

Depends on https://github.com/mirage/ocaml-conduit/pull/369 to avoid conflicts.


samoht:

PR #372, opened 2020-12-08
[WIP] Remove endpoint

dinosaure:
It seems that nobody found a consensus about an ideal definition of an `endpoint`.

To be clear, the goal of `Conduit`, the final result, is: how I can let the user to initialise a "connection" into one and unique shape? The problem is complex because `Conduit` does not have any clues about which protocols are available (even TCP/IP), which _form_ they have (`mirage-tcpip`, `Lwt_unix.file_descr`, `Unix.file_descr`?) and how we initialise them.

In other words and more pragmatically, we want:
```ocaml
val send : flow -> string -> int
val recv : flow -> bytes -> int or_eof
val close : flow -> unit
val resolve : 'a -> flow
```

In Conduit 2.0, a choice was made to say that `'a` is an `Uri.t`, however:
- you can not _decorate_ such value with some others _metadata_ (such as the TLS configuration)
- it does not really fit for Git which accepts `git@github.com` (it's not a valid `Uri.t`)
- Even if you can express lot of things with `Uri.t`, we reach limitations when we talk about TLS, SSH or some others fancy protocols

So a decision was made to use an `Endpoint.t` but we finally come back to the same situation with the same question: `Endpoint.t` is enough to expression **any** endpoints? However, we did not redo the wheel because, instead of Conduit 2.0, we are able to _decorate_ this `Endpoint.t` with some others user-defined _metadata_.

But let's come back a little in the history and see that Conduit 3.0 (or the Tuyau) delivered a solution such as we can provide:
```ocaml
val resolve : 'a t -> 'a -> flow
```
This version was able to let the user to use its own value (SSH private key, TLS configuration, an `Unix.sockaddr`) to initialise a connection. Terms used in this version was not clear but it corresponds to an idea from @samoht: shoud we use extensible variant to define _endpoints_? And yes, we already do that but we never expose such things.

This PR wants to come back to the old version of Conduit (e.g. Tuyau) & re-interprets types and functions to a more comprehensible way - I hope.

### A value

Now, Conduit as a new type (which does not differ from the old type `type 'a key`):
```ocaml
type 'a value
```

This type permits to create a _type witness_ which will represents a value. The user is able to let Conduit to create such type witness with the register function:
```ocaml
val register : name:string -> (module PROTOCOL with type flow = 'flow and type endpoint = 'edn) -> 'edn value * ('edn, 'flow) protocol
```

Or he can create it by himself and re-use it to register his protocol:
```ocaml
val register_with : 'edn value -> (module PROTOCOL with type flow = 'flow and type endpoint = 'edn) -> ('edn, 'flow) protocol
```

Such type witness permits to the **end**-user to fill a `context` which is like our current `resolvers` but instead to keep only functions (`resolver`), it can store values and functions. The idea is to use such `context` to store informations such as the domain-name, the port, the TLS configuration, etc. from the user-context.

Plus, this context is able to store some function such as some resolution processes (like DNS resolver). From the knowledge of what the user is able to inform to Conduit (by the `context` value), we can automatically launch some processes which are some requirements to resolve some values such as the `Unix.sockaddr`. Of course, and again, such function depends on the user context again (if he wants to use `ocaml-dns`, `Unix.gethostbyname` or any others resolution mechanisms).

In others words, from the point of view the user, the goal is to fill this `context` with all available information like:
```ocaml
let ctx = Conduit.empty

let connect uri =
  let ctx = Option.fold ~default:ctx (Conduit_lwt.add Conduit_lwt.TCP.port ctx) (Uri.port uri) in
  let ctx = Option.fold ~default:ctx (Conduit_lwt.add Conduit_lwt.TCP.domain_name ctx) (Uri.host uri) in
  let ctx = Conduit_lwt.TCP.resolver ctx in
  Conduit_lwt.resolve ctx >>= fun flow -> ...
```

In this example, the 2 first lines insert a value to the context and the third line add a DNS resolver. This DNS resolver requires some _metadata_: the domain-name and the port and it take them from the `ctx`. Of course, we can extend `connect` with `?(ctx= ctx)` and prioritize values introduced by the end-user instead to directly use those from the given `Uri.t` (a `context` is just a `Map.S`, so we can use `add`/`replace`, etc.)

Such way seems to fit nicely with MirageOS and Functoria when we can use a simple function such as:
```ocaml
let conduit_ssh_conf conduit =
  let packages_v = [ package "awa-ssh" ] in
  let keys = conduit_ssh_key @?? [] in
  let connect _i modname = function
    | [ conduit ] -> Fmt.strf "let ctx = Conduit_mirage.add SSH.key %s %s.ctx" pp_key ssh_key conduit
    | _ -> Fmt.strf "Conduit.empty" in
  impl ~keys ~connect [ conduit @-> conduit ]
```

Note that it's an idea but I really think that it should be possible to pass values from the CLI to the `Conduit.context`.

### The context

So as you can imagine, the `Conduit.context` can store anything and this the new `Endpoint.t` finally, we do now:
```ocaml
val resolve : context -> flow
```

In that way, I added several convenience functions to fill the context with _metadata_ available on the end-user context such as:
```ocaml
val port : int -> context -> context
val domain_name : [ `host ] Domain_name.t -> context -> context
val dns_resolver : context -> context
```

A problem persists however in this view when the given `context` already have needed values to launch a protocol. For example, we can believe that the user already filled the context with an Unix domain name but a port and and domain-name (with a DNS resolver) exists too. In that case we are able to provide 2 `Unix.sockaddr` and we must prioritize one to the other. I deleted the prioritization mechanism (and tests about that), it should be easy to re-implement it but I did not think about that deeply.

The new design (comparing the current status of `Conduit`) is that the context can be filled by:
- a simple value (and logically, any protocol registration produce a _type witness_ of these values)
- a function

The resolvers still is local (a simple `Map.S`) because even if we are aware about protocol available in our contexts, the user can decide to launch a connection to several ways (one with TLS, one without for example or with a specific configuration). So the end-user must keeps the control on that.

I integrated the PR from @CraigFe which seems nice on this PR.

If you have any question, I will happy to explain details. Then, it seems hard to finally find a consensus on Conduit about several actors so this is my last PR and contribution on this project.


samoht:
I've just switched the master branch and the `v3` branch. If you are still interested by experimenting on the v3 API, would you mind re-opening a PR vs the v3 branch?

PR #373, opened 2021-01-29
[v2] tooling refresh

samoht:
- Update to dune>2
- Use ocamlformat
- remove compilation warnings
- remove travis scripts


samoht:
Note: I've created a `v2` branch on mirage/ocaml-conduit to continue the maintenance of the v2 API of conduit.

PR #374, opened 2021-01-30
more cleanups: reorg files and compile tests

samoht:
Only compile the lwt-unix and async tests as the mirage tests are a bit harder to compile. Also we do not run the tests -- might be worth cherry-picking the changes that happened on `master` for this if really necessary.


hannesm:
This PR introduced a module `X509` in Conduit_lwt_tls.real.mli (and ml) -- is there a reason for this? can we remove it (is it already used elsewhere)?

samoht:
What does it break? Happy to revert it if that's an issue

hannesm:
tls 0.13.0 removed the type aliases (X509_lwt.priv), since that tuple type was revised (now that ECDSA certificates and keys are possible) - and I didn't see the value of an alias to the `Tls.Config.certchain` type.

samoht:
is there any other aliases that can removed in https://github.com/mirage/ocaml-conduit/blob/master/src/conduit-lwt-unix/conduit_lwt_tls.real.mli#L20-L26 ? What about `X509.Authenticator.t` ? I think I've just added this because I didn't know it was an alias :-) feel free to change to something better!

PR #375, opened 2021-01-30
tls: add client-side CA authenticator

samoht:


hannesm:
this is good imho, though two nits:
- it should be documented (I'm not sure in which scenarios conduit is used, but previously no certificate was validated -- now it is validated against the OS trust anchors -- this works for the public internet, but won't for private network setups with your custom CA)
- I'd appreciate a corresponding change to conduit-mirage (I briefly looked into the details, and it seems that getting a Mirage_clock.PCLOCK (for ca-certs-nss / Ca_certs_nss.Make) from within conduit_mirage is rather painful (will need API / functor changes, which need to be reflected in the mirage utility)

TL;DR: best document that now OS trust anchors are used (and wait if someone has a usecase where they need their own (or no) trust anchors) :) and revising conduit_mirage would be appreciated.

samoht:
Updated with changes to `conduit-mirage` too. It seems that only `mirage-skeleton` is using client-side conduit, so I can fix that after the release.

samoht:
Conduit_mirage is really terrible :p maybe it's a good time to simplify this again (but in a different PR).

src/conduit-lwt-unix/conduit_lwt_unix.mli
```diff
@@ -166,7 +166,8 @@ val init :
 (** [init ?src ?tls_own_key ()] will initialize a Unix conduit that binds to the
     [src] interface if specified. If TLS server connections are used, then
     [tls_server_key] must contain a valid certificate to be used to advertise a
-    TLS connection *)
+    TLS connection. The certificate is validated against the OS trust anchors
```
avsm, line 169: Can't we pass through the `authenticator` function here as an optional arg to init? That would allow it to be overridden for custom CAs, but the default would work for public anchors.

samoht:
@avsm I've pushed a commit to select the correct authenticator.

I am not totally sure how to do this with `ocaml-ssl` so it's TLS-only.

src/conduit-lwt-unix/conduit_lwt_unix.mli
```diff
@@ -160,13 +160,16 @@ val default_ctx : ctx
...
   unit ->
   ctx io
 (** [init ?src ?tls_own_key ()] will initialize a Unix conduit that binds to the
     [src] interface if specified. If TLS server connections are used, then
     [tls_server_key] must contain a valid certificate to be used to advertise a
-    TLS connection *)
+    TLS connection.
+
+    The certificate is validated using [tls_authenticator]. By default, the
+    validation is using the OS trust anchors. *)
```
hannesm, line 172:
```suggestion
    validation is using the {{:https://github.com/mirage/cacerts}OS trust anchors}. *)
```

src/conduit-mirage/conduit_mirage.mli
```diff
@@ -93,8 +93,15 @@ type server = [ tcp_server | vchan_server | server tls_server ]
...
 
-val client : Conduit.endp -> client Lwt.t
-(** Resolve a conduit endpoint into a client configuration. *)
+module Client (P : Mirage_clock.PCLOCK) : sig
+  val resolve :
+    ?tls_authenticator:X509.Authenticator.t -> Conduit.endp -> client Lwt.t
+  (** Resolve a conduit endpoint into a client configuration.
+
+      The certificate is validated using [tls_authenticator]. By default, the
+      validation is using the trust anchors extracted from Mozilla's NSS
```
hannesm, line 102:
```suggestion
      validation is using the {{:https://github.com/mirage/ca-certs-nss}trust anchors extracted from Mozilla's NSS}.
```

src/conduit-mirage/conduit_mirage.mli
```diff
@@ -93,8 +93,15 @@ type server = [ tcp_server | vchan_server | server tls_server ]
...
+      certdata.txt package. *)
```
hannesm, line 103:
```suggestion
```

hannesm:
Thanks @samoht -- I think this is good (moving conduit-lwt and conduit-mirage on the same level).

src/conduit-lwt-unix/conduit_lwt_unix.mli
```diff
@@ -160,13 +160,17 @@ val default_ctx : ctx
...
   unit ->
   ctx io
 (** [init ?src ?tls_own_key ()] will initialize a Unix conduit that binds to the
     [src] interface if specified. If TLS server connections are used, then
     [tls_server_key] must contain a valid certificate to be used to advertise a
-    TLS connection *)
+    TLS connection.
+
+    The certificate is validated using [tls_authenticator]. By default, the
+    validation is using the {{:https://github.com/mirage/cacerts} OS trust
```
avsm, line 172:
```suggestion
    validation is using the {{:https://github.com/mirage/ca-certs} OS trust
```

avsm (approved):
lgtm

avsm:
(needs a changes entry)

samoht:
I've just added a CHANGE entry

PR #376, opened 2021-02-01
[v2] conduit-mirage: simplify the API

samoht:
Do not mix first-class modules and functors anywmore as it makes the
API quite complicated for not much gains: the code is generated by
the mirage tools anyway.

Note: this needs some changes in the mirage tool too. Will do this next if people are generally happy with this simplification.

### DIFF
--- a/conduit-mirage.opam
+++ b/conduit-mirage.opam
@@ -27,6 +27,7 @@ depends: [
   "ca-certs-nss"
   "ipaddr" {>= "3.0.0"}
   "ipaddr-sexp"
+  "tcpip" {with-test}
 ]
 conflicts: [
   "mirage-conduit"
--- a/src/conduit-mirage/conduit_mirage.ml
+++ b/src/conduit-mirage/conduit_mirage.ml
@@ -31,173 +31,68 @@ let err_vchan_not_supported = fail "%s: VCHAN is not supported"
 let err_unknown = fail "%s: unknown endpoint type"
 let err_ipv6 = fail "%s: IPv6 is not supported"
 
-module Flow = struct
-  type error = [ `Msg of string ]
-  type write_error = [ Mirage_flow.write_error | error ]
+let err_not_supported = function
+  | `TLS _ -> err_tls_not_supported
+  | `TCP _ -> err_tcp_not_supported
+  | `Vchan _ -> err_vchan_not_supported
 
-  let pp_error ppf (`Msg s) = Fmt.string ppf s
+type client =
+  [ `TCP of Ipaddr_sexp.t * int
+  | `TLS of Tls.Config.client * client
+  | `Vchan of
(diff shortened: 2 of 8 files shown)


samoht:
@hannesm is this the right place to initialise the RNG device?

I've tested this with https://github.com/mirage/ocaml-cohttp/pull/742 and https://github.com/mirage/ocaml-cohttp/pull/742 and mirage-skeleton compiles (and the examples seem to run) with https://github.com/mirage/mirage-skeleton/pull/295

hannesm:
@samoht
> is this the right place to initialise the RNG device?

no, the mirage tool itself -- when someone depends on "random" / "default_random" -- emits the initialization sequence during boot (see https://github.com/mirage/mirage/blob/b121300fa2bc17c80d42283ac7e4d2ffe4542fd0/lib/mirage_impl_random.ml#L20) -- as long as you're using the `mirage` utility (if not, please depend on mirage-crypto-rng-mirage and call `Mirage_crypto_rng_mirage.initialize ()` in the same way mirage does it). Using Mirage_crypto_rng_lwt in conduit-mirage results in (unavailable) calls to the `getrandom` system call (and thus should be avoided).

For conduit-lwt, which uses tls.lwt, there's no need to invoke the Mirage_crypto_rng_lwt.initialize (), since tls_lwt already does it.

dinosaure:
I follow @hannesm when `mirage-crypto-rng.lwt` depends on `lwt.unix`. So, transitively, we break any MirageOS with Solo5.

samoht:
Odd -- if I don't add the RNG initialialisation function here, some mirage-skeleton examples complain that it is missing. I'll double check to see what happened.

samoht:
> For conduit-lwt, which uses tls.lwt, there's no need to invoke the Mirage_crypto_rng_lwt.initialize (), since tls_lwt already does it.

@hannesm so I don't understand https://github.com/mirage/ocaml-conduit/commit/175ab18dfa9c52f2a735c0b7f8a81fad66ca8422 then

hannesm:
@samoht me neither, since tls 0.12.2 this is done in tls.lwt (ocamlfind) Tls_lwt (module): https://github.com/mirleft/ocaml-tls/commit/46da4b8a91e398fc09f3d0663156b3b9d169b6c1

the initialization in Mirage_crypto_rng_lwt.initialize (https://github.com/mirage/mirage-crypto/blob/main/rng/lwt/mirage_crypto_rng_lwt.ml#L39) is idempotent, so it doesn't matter whether it is called in conduit as well. but there's no need to do so.

samoht:
The CI is currently broken due to https://github.com/ocaml/opam-repository/pull/18071 -- it's unclear to me how we should fix this but I'll try a fix in opam-repo.

samoht:
(see https://github.com/ocaml/opam-repository/pull/18090 for a tentative fix)

PR #379, opened 2021-03-22
Make listening port discoverable

torinnd:
There's currently no way to learn what port a Conduit v3 Async server is listening on. Interrogating a [Tcp.Where_to_listen.t] doesn't always include accurate information, specifically in the special case of [Where_to_listen.of_port_chosen_by_os].

I'd like to propose introducing an optional [listening_on] Ivar.t that will be filled with the relevant port information from the bound [Socket.t] after `main ()` is called.

### DIFF
--- a/src/async-ssl/conduit_async_ssl.ml
+++ b/src/async-ssl/conduit_async_ssl.ml
@@ -302,8 +302,8 @@ module TCP = struct
   let protocol =
     protocol_with_ssl ~reader:Protocol.reader ~writer:Protocol.writer protocol
 
-  let service =
-    service_with_ssl service ~reader:Protocol.reader ~writer:Protocol.writer
+  let service ?listening_on () =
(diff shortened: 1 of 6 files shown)


dinosaure:
`master` is not anymore the version 3.0.0 of Conduit. It should be better to compare what you did with the `v3` branch. Then, I'm not sure about the next step.

torinnd:
Yeah, sorry about that. I've fixed the base branch.

avsm:
Are you using Conduit.3.0.0 in an application @torinnd?  Do you specifically need conduit-3 or can you use 2 or 4 instead?

torinnd:
I'm not using Conduit 3.0.0 in any real-world applications. I was just trying to reconcile the proposed interface with existing code that relied on Conduit 2.0.0. I'll close this PR.

avsm:
Thanks (and your contribution is much appreciated!). We'll be revisiting the conduit 3 interface just as a few other things settle down around the MirageOS 4 release.

PR #380, opened 2021-03-27
Update conduit-mirage to v4v6

dinosaure:
According to the current state of `dns-client` (/cc @hannesm), it's a simple patch to upgrade `conduit` to `v4v6` stack. Currently, we are a bit stuck with: https://github.com/mirage/mirage/pull/1209 which must be merged to upgrade mirage conduit device then to `v4v6`.

### DIFF
--- a/conduit-mirage.opam
+++ b/conduit-mirage.opam
@@ -38,5 +38,8 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest" "-p" name] {with-test}
 ]
+pin-depends: [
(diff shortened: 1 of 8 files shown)


conduit-mirage.opam
```diff
@@ -38,5 +38,8 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
```
hannesm, line 38: above this needs a `"mirage-stack" {>= "2.2.0"}` constraint.

dinosaure:
Ready to merge (the error about `alpine` is referenced to savonet/ocaml-ssl#70).

PR #385, opened 2021-04-19
ipaddr-sexp in conduit-async

anmonteiro:
`conduit-async` is missing a dependency on `ipaddr-sexp`.

not sure why this is needed by @sgrove couldn't compile `conduit-async` without it.


dinosaure:
Rebased on `master` due to the new constraint on `dns.5.0.0`, seems ok to merge, Thanks :+1:!

PR #387, opened 2021-04-22
conduit-lwt-unix: se accept_n on the server

mseri:
In benchmarks for cohttp-lwt-unix' latency we see evidence of connections starvation.
Using `accept_n` seems to improve the situation. You can see [here](https://gist.github.com/mseri/46bee49f50a46267753cfba9d0c24571) an example of the artificial benchmarks that I used to check this

Ping @avsm

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_server.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_server.ml
@@ -71,34 +71,33 @@ let run_handler handler v =
                   f "Uncaught exception in handler: %s" (Printexc.to_string ex)));
           Lwt.return_unit))
 
+let log_exn = function
+  | Some ex ->
+      Log.warn (fun f ->
+          f "Uncaught exception accepting connection: %s"
+            (Printexc.to_string ex))
+  | None -> ()
+
 let init ?(stop = fst (Lwt.wait ())) handler fd =
   let stop = Lwt.map (fun () -> `Stop) stop in
   let rec loop () =
-    Lwt.try_bind
-      (fun () ->
-        connected ();
-        throttle () >>= fun () ->
-        let accept = Lwt.map (fun v -> `Accept v) (Lwt_unix.accept fd) in
-        Lwt.choose [ accept; stop ] >|= function
-        | `Stop ->
-            Lwt.cancel accept;
-            `Stop
-        | `Accept _ as x -> x)
-      (function
-        | `Stop ->
-            disconnected ();
-            Lwt.return_unit
-        | `Accept v ->
+    let accepted =
+      Lwt_unix.accept_n fd 10000 >>= fun (connections, maybe_error) ->
+      log_exn maybe_error;
+      Lwt.return (`Accepted connections)
+    in
+    Lwt.choose [ accepted; stop ] >>= function
+    | `Stop ->
+        Lwt.cancel accepted;
+        Lwt.return_unit
+    | `Accepted connections ->
+        Lwt_list.iter_p
+          (fun v ->
+            connected ();
+            throttle () >>= fun () ->
             run_handler handler v;
-            loop ())
-      (fun exn ->
-        disconnected ();
-        match exn with
-        | Lwt.Canceled -> Lwt.return_unit
-        | ex ->
-            Log.warn (fun f ->
-                f "Uncaught exception accepting connection: %s"
-                  (Printexc.to_string ex));
-            Lwt_unix.yield () >>= loop)
+            Lwt.return_unit)
+          connections
+        >>= fun () -> loop ()
   in
   Lwt.finalize loop (fun () -> Lwt_unix.close fd)
--- a/src/conduit-lwt-unix/conduit_lwt_server.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_server.ml
@@ -71,7 +71,7 @@ let run_handler handler v =
                   f "Uncaught exception in handler: %s" (Printexc.to_string ex)));
           Lwt.return_unit))
 
-let init ?(stop = fst (Lwt.wait ())) handler fd =
+let init ?(nconn=10_000) ?(stop = fst (Lwt.wait ())) handler fd =
   let stop = Lwt.map (fun () -> `Stop) stop in
   let log_exn = function
     | Some ex ->
@@ -82,7 +82,7 @@ let init ?(stop = fst (Lwt.wait ())) handler fd =
   in
   let rec loop () =
     let accepted =
-      Lwt_unix.accept_n fd 10000 >>= fun (connections, maybe_error) ->
+      Lwt_unix.accept_n fd nconn >>= fun (connections, maybe_error) ->
       log_exn maybe_error;
       Lwt.return (`Accepted connections)
     in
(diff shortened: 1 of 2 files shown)


src/conduit-lwt-unix/conduit_lwt_server.mli
```diff
@@ -10,6 +10,7 @@ val process_accept :
   unit Lwt.t
 
 val init :
```
mseri, line 12: I should probably document this, I was first waiting to see if you are interested at all and what the review says about the implementation

mseri:
Adding the yield call greatly improves the latency, but apparently at a cost on lost connections
latency

src/conduit-lwt-unix/conduit_lwt_server.ml
```diff
@@ -71,34 +71,41 @@ let run_handler handler v =
...
-        | `Accept v ->
-            run_handler handler v;
-            loop ())
-      (fun exn ->
-        disconnected ();
-        match exn with
+        | `Accepted connections ->
+            Lwt_list.iter_p
+              (fun v ->
+                connected ();
```
mseri, line 98: There may also be a race to update connected here since I am calling `iter_p`, right?

mseri:
I have updated the test, using the test server suggested in https://github.com/inhabitedtype/httpaf/issues/200
The new results are in https://gist.github.com/mseri/d40130fef7c8673e55b9e337f411c56d and summarised below

latency

In practice seems that using `accept_n` does improve the latency with respect both to the current implementation and the one with the extra yield call

src/conduit-lwt-unix/conduit_lwt_server.ml
avsm: I don't quite see the race you're referring to here -- where is the `iter_p`?

avsm (approved):
This looks like an improvement to me for sure. I'd like to run tests on Linux before merging it just to be quantify the win.

src/conduit-lwt-unix/conduit_lwt_server.ml
mseri: If I am not mistaken `iter_p` runs the iteration in parallel, and `connected ()` increments the active connection count without any protection (`let connected () = incr active`). My fear was that multiple connections executed in parallel could try to increment the variable simultaneously (and I don't want to put a mutex there).

mseri:
If somebody can benchmark it on linux, can you also check what difference it does with and without https://github.com/mirage/ocaml-conduit/pull/387/commits/cf7c62f6c3c6edc9eb3686541fadb27800d47c65

anuragsoni:
> @mseri  If somebody can benchmark it on linux, can you also check what difference it does with and without [cf7c62f](https://github.com/mirage/ocaml-conduit/commit/cf7c62f6c3c6edc9eb3686541fadb27800d47c65) ?

I gave this a run on linux today. I did a set of different tests. I first tried running the tests through cohttp + conduit and took the example server from your gist. I didn't notice a big difference between the current released conduit and the version with accept_n. I then tried to use conduit for my own WIP http 1.1 server implementation and tried the following combinations:

All servers run a service of the form `req -> res Lwt.t`
* Server with upstream conduit where handler returns ASAP
* Server with accept_n where handler returns ASAP
With these two servers there wasn't a big difference between accept_n and accept.

* Server with upstream conduit where handler uses `Lwt.pause` before responding
* Server with accept_n where handler uses `Lwt.pause` before responding
In these two servers I noticed improvement in the accept_n branch vs the current released conduit.

All of the data can be found at https://gist.github.com/anuragsoni/33ae4a750dd7e110cbe94ed69614b7d5
Do let me know if something looks missing there and i'd be happy to run more tests.

As an aside (and this is most likely off-topic for conduit and maybe more a discussion about how lwt scheduler works?), another potential data point was when i tried running a test that can simulate handlers that do a little bit of work. I added a sleep of 1 millisecond in the handler before it responds. With this change i noticed a big spike in latency and the numbers went up from milliseconds to seconds. https://gist.github.com/anuragsoni/33ae4a750dd7e110cbe94ed69614b7d5#file-h1_conduit_accept_n_1ms_delay-txt

I've been running some benchmarks against servers in other languages and i don't see a similar spike when i add a millisecond delay in a handler for a tokio based rust web server, and I don't see a similar spike when I run my `h1` server on async instead of lwt. The rust and h1-async results can be seen at https://github.com/anuragsoni/h1/tree/3c16f5eccf4a17c1009814a5178537fb50e5aaad/wrk_bench
I'm still early in my experiments so I'm not a 100% sure about what causes these spikes for Lwt, but i'll spend some more time to see if I can get more information about what's happening!

mseri:
Dear @anuragsoni thanks a lot for this!

The idea of adding the millisecond pause is brilliant.

There are a number of tickets on this latency issue, if would be great if we can figure out how to improve/fix it.

anuragsoni:
> @mseri  There are a number of tickets on this latency issue, if would be great if we can figure out how to improve/fix it.

For sure! I remember seeing some tickets on cohttp and httpaf, are those the tickets you are referring to? 

I'm wondering if some of the latency related investigation/work will be needed at a layer below that? I've only recently started using Lwt seriously, but the fact that I'm seeing the large latency spike only on lwt and not on async leads me to believe that it'd be worthwhile to see if we can come up with a small reproduction case that's independent of cohttp/conduit etc to see if the issue lies somewhere in Lwt. When i tested the 1ms delay on async with my h1 library I saw a similar spike in latency when i was using the higher level Reader/Writer modules that do their own buffering and scheduling for writes. I noticed pretty significant improvements when i switched to using async's file_descr/socket apis directly to orchestrate my own read/writes and do my own buffering. I was hoping to see similar improvements on my lwt backend and I do see improvements when the handler doesn't do any work, but with the 1ms delay going down to read/write calls with Lwt_unix.file_descr's still shows spikes in latency that are similar to when I was testing with the Lwt_io managed buffered channels.

Do note that everything I said might very well be wrong as I'm still exploring the lwt api and learning more about it :D

mseri:
What you see is reasonable and in line with some of the comments in the tickets I am referring to. Give me a day and I will link them here

mseri:
Here we go, the latency issue is discussed in

    - Latency benchmarking cohttp shows queueing [cohttp#328](https://github.com/mirage/ocaml-cohttp/issues/328)
    - Memory leak under heavy load [cohttp#545](https://github.com/mirage/ocaml-cohttp/issues/545)
    - "Hello world" HTTP server dies under load  [cohttp#503](https://github.com/mirage/ocaml-cohttp/issues/503), possibly what happens in [cohttp#412](https://github.com/mirage/ocaml-cohttp/issues/412)

as well as [httpaf#200](https://github.com/inhabitedtype/httpaf/issues/200) and [eliom#270](https://github.com/ocsigen/eliom/pull/670) -- ping also @smorimoto that made some tests there --

It can be mitigated a bit manipulating
```ocaml
    Conduit_lwt_unix.set_max_active 100;
    Lwt_io.set_default_buffer_size 0x10000
```
and, as you observed, by using `accept_n` and yielding there.
There may also be a relation to [Possible memory leak in client.read_response cohttp#688](https://github.com/mirage/ocaml-cohttp/issues/688) – which, in fact, goes hand in hand with another issue in Lwt, closed for lack of manpower.

mseri:
There is a new benchmark showing how httpaf crushes cohttp on unilernels: https://github.com/roburio/unipi/pull/4#issuecomment-870923047

PR #389, opened 2021-04-23
Allow openssl conduit-lwt-unix users to control their client Ssl.contexts

lippirk:
I am making this patch because I'd like to be able to control the
lifetime of my openssl client context directly. 

Motivation for this includes the fact that the global
`Conduit_lwt_unix_ssl.Client.default_ctx` is not sufficient in the case
that you want to update an ssl context's verify location (it is
necessary to use a new ssl context).

Example usage with cohttp:

```
(* three types of context :D *)
let client_ssl_context : Ssl.context = Conduit_lwt_unix_ssl.Client.create_ctx () in
(* some ssl setup *)
let* (ctx : Conduit_lwt_unix.ctx) = Conduit_lwt_unix.init ~client_ssl_context () in
let ctx : Cohttp_lwt_unix.Client.ctx = Cohttp_lwt_unix.Client.custom_ctx ~ctx () in
let* _ = Cohttp_lwt_unix.Client.call ~ctx `POST ~body:"blah" uri in
...
```

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_unix.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix.ml
@@ -108,6 +108,7 @@ type ctx = {
   src : Unix.sockaddr option;
   tls_own_key : tls_own_key;
   tls_authenticator : Conduit_lwt_tls.X509.authenticator;
+  client_ssl_context : Ssl.context option ;
 }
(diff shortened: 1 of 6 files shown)


lippirk:
closing in favour of #390

PR #390, opened 2021-04-26
Add openssl_overrides to conduit context

lippirk:
This patch really belongs in cohttp, but I don't think conduit
exposes all the right machinery.

This patch allows openssl cohttp-lwt-unix users to:
(a) connect to a particular hostname/IP, but specify something else to
    verify against
(b) have direct control over the lifetime of their client's ssl context

I believe you should be able to achieve (a) using the following
resolver (but it results in 'not supported' errors):

```
let resolver =
  let table = Hashtbl.create 16 in
  let cn = "expected-cn" in
  let Ok ip = Ipaddr.of_string cn in
  Hashtbl.add table cn (`TLS (cn, `TCP (ip, port)));
  Resolver_lwt_unix.static table
```

It's not possible to achieve (b) right now (at least in v2).

(b) is useful if your trusted bundle changes (calling
`load_verify_locations` with the same SSL context does not work as
one might expect). The alternative is to restart your application.

Intended usage with cohttp:

```
let ctx : Ssl.context = ... in
let openssl_overrides =
  let open Conduit_lwt_unix_ssl.Overrides in
  {
    client =
      Some Client.{ ctx = Some ctx; hostname = Some cn };
  }
in
let* (ctx : Conduit_lwt_unix.ctx) = Conduit_lwt_unix.init ~openssl_overrides () in
let ctx : Cohttp_lwt_unix.Client.ctx = Cohttp_lwt_unix.Client.custom_ctx ~ctx () in
let* _resp, resp_body = Client.call ~ctx `POST ~body uri in
...
```

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_unix.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix.ml
@@ -108,6 +108,7 @@ type ctx = {
   src : Unix.sockaddr option;
   tls_own_key : tls_own_key;
   tls_authenticator : Conduit_lwt_tls.X509.authenticator;
+  openssl_overrides : Conduit_lwt_unix_ssl.Overrides.t option;
 }
 
 let string_of_unix_sockaddr sa =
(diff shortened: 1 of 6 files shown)


lippirk:
I resolved some conflicts to do with the lazy changes

lindig:
How are the changes of getting this merged? We are also considering a refinement of this to support not verifying the hostname. This is a behavior that `openssl s_client` implements but that currently can't be configured using conduit.

psafont:
This can now be closed now that we've settled on a different method to bypass hostname verification and pass the ssl context to the OpenSSL clients

Issue #388, opened 2021-04-22
Conduit crashes on startup due to ca-certifcates, even when not using tls


talex5:
```
$ utop
utop # #require "cohttp-lwt-unix";;
Exception:
(Failure
   "ca-certs: no trust anchor file found, looked into /etc/ssl/certs/ca-certificates.crt, /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem, /etc/ssl/ca-bundle.pem.\
  \nPlease report an issue at https://github.com/mirage/ca-certs, including:\
  \n- the output of uname -s\
  \n- the distribution you use\
  \n- the location of default trust anchors (if known)\
  \n")
Error: Reference to undefined global `Conduit_lwt_unix'
```
Probably this should be `lazy`: https://github.com/mirage/ocaml-conduit/blob/a2f91d0fb17c9b0a8436dd39c4267b9144f0bdea/src/conduit-lwt-unix/conduit_lwt_tls.real.ml#L25-L28

(I hit this when I updated ocluster-scheduler. It doesn't use https, so the Docker images didn't contain any certificates.)

PR #395, opened 2021-05-21
Search for TLS trust anchors lazily

CraigFe:
Changes the type of `Conduit_lwt_tls.X509.default_authenticator` and `Conduit_lwt_unix.default_ctx` to be lazy, avoiding various side-effects (system interactions, logging) due to constructing these values at initialisation time.

Fixes https://github.com/mirage/ocaml-conduit/issues/388.


avsm:
This'll require a corresponding PR in cohttp-lwt to do the Lazy.force, and then both can be released together.

PR #396, opened 2021-08-02
conduit-lwt-unix: adapt to tls 0.14.0 API

hannesm:
https://github.com/ocaml/opam-repository/pull/19175

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_unix.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix.ml
@@ -264,6 +264,15 @@ let connect_with_tls_native ~ctx (`Hostname hostname, `IP ip, `Port port) =
       Conduit_lwt_tls.X509.private_of_pems ~cert ~priv_key
       >|= fun certificate -> Some (`Single certificate))
   >>= fun certificates ->
+  let hostname =
+    try Domain_name.(host_exn (of_string_exn hostname))
+    with Invalid_argument msg ->
+      let s =
+        Printf.sprintf "couldn't convert %s to a [`host] Domain_name.t: %s"
+          hostname msg
+      in
+      invalid_arg s
+  in
   Conduit_lwt_tls.Client.connect ?src:ctx.src ?certificates
     ~authenticator:ctx.tls_authenticator hostname sa
   >|= fun (fd, ic, oc) ->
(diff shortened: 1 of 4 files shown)


hannesm:
anyone up for a quick review (+merge +release) to unlock the ecosystem with the new tls version? When looking at the code, I'm slightly confused why there is a multitude of constructors (Tls, Tls_native, OpenSSL), all using the same tls_server_config/tls_client_config. I intended to use the `` [`host] Domain_name.t `` at the API level, but then was unsure about (a) clients of the API and (b) what the OpenSSL library expects. This change is API compatible AFAICT.

src/conduit-lwt-unix/conduit_lwt_unix.ml
```diff
@@ -264,6 +264,15 @@ let connect_with_tls_native ~ctx (`Hostname hostname, `IP ip, `Port port) =
...
   >>= fun certificates ->
+  let hostname =
+    try Domain_name.(host_exn (of_string_exn hostname)) with
+      Invalid_argument msg ->
+      let s =
+        Printf.sprintf "couldn't convert %s to a [`host] Domain_name.t: %s"
+          hostname msg
+      in
+      invalid_arg s
+  in
```
dinosaure, line 275: You probably should update `Conduit_lwt_unix_ssl.real.ml{,i}` and use `Domain_name.to_string` in this place instead of raising an error. I think it's fine to break the API (and require an `[ host ] Domain_name.t` instead of a simple `string` as the host) and be more restrictive even if we want to use OpenSSL. WDYT @avsm?

hannesm: As explained above, I'd leave the API change for someone in the future. I still don't have any idea what "The `conduit` library takes care of establishing and listening for TCP and SSL/TLS connections for the Lwt and Async libraries." means in respect to error handling. Code-wise since `connect` raises, I would find it obscure if only the Domain_name.host / Domain_name.of_string would result in an error (instead I'd expect either "no exceptions are raised" or "no result type is used") -- and the current code is "no result type is used, exceptions are raised"

dinosaure:
To avoid any conflicts for further releases, I will cut a minor release with your patch. Do you agree @avsm?

dinosaure:
I will do a `4.0.1` release without #395 to unlock the dependency graph with `tls.0.14.0`. However, we should have, as soon as we can, `conduit.4.1.0` with, at least, #395 (/cc @CraigFe). I don't know if this PR is urgent or not.

PR #399, opened 2021-08-06
Prepare v4.0.1

dinosaure:
/cc @CraigFe, @hannesm & @avsm. It's a minor release which does not integrate #395 - So I will not merge this PR but I will just produce the tag and the tarball needed for the release. The minor release is needed because:
1) I'm not sure that we want a release which update the interface (and imply several others releases by cascading)
2) Some PR remains and should be a part of the next release (specially #319 & #390)

I will cherry-pick the commit which has the changelog update.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,5 @@
 ## Unreleased
 
-* change the type of `Conduit_lwt_tls.X509.default_authenticator` and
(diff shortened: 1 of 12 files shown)


hannesm:
Thanks @dinosaure, this looks good to me.

dinosaure:
The failure on `ocaml-ci` is unrelated and it's about `core`/`async`. Let's do the release.

PR #400, opened 2021-08-08
Revert "Prepare v4.0.1"

dinosaure:
Reverts mirage/ocaml-conduit#399


hannesm:
I think we should re-apply 38b5fe6 in this PR (or avoid the revert of that commit).

dinosaure:
ah yes, I will do.

PR #410, opened 2021-12-11
conduit-mirage uses Tcpip_stack_socket.V4 and V4V6 - which are only available since tcpip 6.0.0

hannesm:


hannesm:
part of #413

PR #414, opened 2022-06-21
conduit-mirage: pass peer name to Tls.Config.client

hannesm:
//cc @palainp this solves the issue you reported at https://github.com/mirleft/ocaml-tls/issues/447 with the http-fetch unikernel.

Dear ocaml-conduit maintainers: what is the proper way to deal with `host` (here being a string) does not conform being a  `` [ `host ] Domain_name.t ``? In this PR, it is silently ignored -- I know not much about conduit and its semantics and usage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


hannesm:
ok, CI is green. any comments @dinosaure @samoht

samoht:
LGTM

PR #415, opened 2022-10-05
Delay parsing of nameservers

reynir:
If we want to use TLS we need to be able to create an authenticator. And many interesting authenticators require ca-certs.


dinosaure (approved):

dinosaure:
Seems relatively a good change which require a new release (we break the API), let's do that next week to unlock the PR on `mirage/mirage`.

PR #417, opened 2022-11-07
conduit-lwt-unix: allow openssl clients to customize the ssl context and the verification

psafont:
Opening the PR to gather comments and discussion around it, supersedes https://github.com/mirage/ocaml-conduit/pull/390

This PR allows clients to have complete control of the ssl context used by conduit, as the previous PR intended. Unlike it, this one integrates into the conduit context creation, like the client-side authenticator. It doesn't allow to control the hostname that verifies the server. xapi doesn't use it anymore and doubt it's going to be useful for anybody else.

It similarly includes into the context a verification control to be able to set whether to verify the hostname and/or the IP

Issues:
- [ ] The name of the new parameter `ssl_ctx` does not make it explicit that it will be used only for client connections, not server ones. If clients want to change the ssl context for server connections a new parameter will be added and might make things confusing. Do we want to use `client_ssl_ctx` here? I can even add `server_ssl_ctx` parameter even if I don't have a need for it.

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.ml
@@ -18,6 +18,7 @@
 module Client = struct
   type verify = { hostname : bool; ip : bool }
 
+  let default_verify = { hostname = true; ip = true }
   let default_ctx = `Ssl_not_available
   let create_ctx ?certfile:_ ?keyfile:_ ?password:_ () = default_ctx
 
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.ml
@@ -16,10 +16,12 @@
  *)
 
 module Client = struct
+  type verify = { hostname : bool; ip : bool }
+
   let default_ctx = `Ssl_not_available
   let create_ctx ?certfile:_ ?keyfile:_ ?password:_ () = default_ctx
 
-  let connect ?(ctx = default_ctx) ?src:_ ?hostname:_ _sa =
+  let connect ?(ctx = default_ctx) ?src:_ ?hostname:_ ?ip:_ ?verify:_ _sa =
     ignore ctx;
     Lwt.fail_with "Ssl not available"
 end
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.mli
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.mli
@@ -20,6 +20,7 @@
 module Client : sig
   type verify = { hostname : bool; ip : bool }
 
+  val default_verify : verify
   val default_ctx : [ `Ssl_not_available ]
 
   val create_ctx :
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.mli
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.mli
@@ -18,6 +18,8 @@
 (** TLS/SSL connections via {{:http://www.openssl.org} OpenSSL} C bindings *)
 
 module Client : sig
+  type verify = { hostname : bool; ip : bool }
+
   val default_ctx : [ `Ssl_not_available ]
 
   val create_ctx :
@@ -31,6 +33,8 @@ module Client : sig
     ?ctx:[ `Ssl_not_available ] ->
     ?src:Lwt_unix.sockaddr ->
     ?hostname:string ->
+    ?ip:Ipaddr.t ->
+    ?verify:verify ->
     Lwt_unix.sockaddr ->
     (Lwt_unix.file_descr * Lwt_io.input_channel * Lwt_io.output_channel) Lwt.t
 end
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
@@ -95,21 +95,18 @@ module Client = struct
         let maybe_verify ssl = function
           | Some hostname, Some ip ->
               Ssl.set_hostflags ssl [ Ssl.No_partial_wildcards ];
+              Ssl.set_client_SNI_hostname ssl hostname;
               Ssl.set_host ssl hostname;
               Ssl.set_ip ssl (Ipaddr.to_string ip)
           | Some hostname, None ->
               Ssl.set_hostflags ssl [ Ssl.No_partial_wildcards ];
+              Ssl.set_client_SNI_hostname ssl hostname;
               Ssl.set_host ssl hostname
           | None, Some ip -> Ssl.set_ip ssl (Ipaddr.to_string ip)
           | None, None -> ()
         in
-        (match hostname with
-        | Some host ->
-            with_socket (fun ssl ->
-                Ssl.set_client_SNI_hostname ssl host;
-                maybe_verify ssl to_verify)
-        | None -> Lwt_ssl.ssl_connect fd ctx)
-        >>= fun sock -> Lwt.return (chans_of_fd sock))
+        with_socket (fun ssl -> maybe_verify ssl to_verify) >>= fun sock ->
+        Lwt.return (chans_of_fd sock))
 end
 
 module Server = struct
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
@@ -54,6 +54,8 @@ module Client = struct
 
   type verify = { hostname : bool; ip : bool }
 
+  let default_verify = { hostname = true; ip = false }
+
   let validate_hostname host_addr =
     try
       let _ = Domain_name.(host_exn (of_string_exn host_addr)) in
@@ -76,7 +78,7 @@ module Client = struct
         (hostname, ip)
 
   let connect ?(ctx = default_ctx) ?src ?hostname ?ip ?verify sa =
-    let verify = Option.value ~default:{ hostname = true; ip = false } verify in
+    let verify = Option.value ~default:default_verify verify in
     let to_verify = verification verify (hostname, ip) in
     Conduit_lwt_server.with_socket sa (fun fd ->
         (match src with
--- a/src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
@@ -52,22 +52,60 @@ module Client = struct
 
   let default_ctx = create_ctx ()
 
-  let connect ?(ctx = default_ctx) ?src ?hostname sa =
+  type verify = { hostname : bool; ip : bool }
+
+  let validate_hostname host_addr =
+    try
+      let _ = Domain_name.(host_exn (of_string_exn host_addr)) in
+      host_addr
+    with Invalid_argument msg ->
+      let s =
+        Printf.sprintf "couldn't convert %s to a [`host] Domain_name.t: %s"
+          host_addr msg
+      in
+      invalid_arg s
+
+  let verification { hostname; ip } = function
+    | None, _ when hostname -> invalid_arg "impossible to verify hostname"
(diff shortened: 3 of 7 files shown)


src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.ml
```diff
@@ -16,10 +16,13 @@
...
 
 module Client = struct
+  type verify = { hostname : bool; ip : bool }
+
+  let default_verify = { hostname = true; ip = true }
   let default_ctx = `Ssl_not_available
   let create_ctx ?certfile:_ ?keyfile:_ ?password:_ () = default_ctx
 
-  let connect ?(ctx = default_ctx) ?src:_ ?hostname:_ _sa =
+  let connect ?(ctx = default_ctx) ?src:_ ?hostname:_ ~ip:_ ~verify:_ _sa =
```
edwintorok, line 25: Please use ?ip and ?verify, looks like the CI may not be verifying this, but if the interface declares optional labeled arguments, the implementations cannot use non-optional ones, even if it ignores them.

psafont: Thanks, I've changed it

psafont:
@hannesm Would you mind reviewing these changes? I've integrated them into a xenserver build and all related tests are passing which gives me confidence it's working. I've also manually tested the defaults for the client and it's verifying the domain name of certificate as previously

hannesm:
I guess this is fine, although it makes the ocaml-tls and openssl versions slightly different. The API is only changed in a minor way (adding optional arguments) which allows old programs to use the new API as well.

src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.mli
```diff
@@ -31,6 +34,8 @@ module Client : sig
     ?ctx:[ `Ssl_not_available ] ->
     ?src:Lwt_unix.sockaddr ->
     ?hostname:string ->
+    ?ip:Ipaddr.t ->
+    ?verify:verify ->
```
hannesm, line 38: I wonder, shouldn't `?verify` be implicit? If`~ip` is provided, likely `ip` in the verify record will be true, if `~hostname` is provided, likely `hostname` in the verify record will be true.

Is there a use case where both `hostname` and `ip` are provided and should be verified? Is there a use case where `hostname` is provided but shouldn't be verified?

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
```diff
@@ -52,22 +52,59 @@ module Client = struct
...
-            let ssl = Lwt_ssl.ssl_socket_of_uninitialized_socket s in
-            Ssl.set_client_SNI_hostname ssl host;
-            (* Enable hostname verification *)
-            Ssl.set_hostflags ssl [ Ssl.No_partial_wildcards ];
-            Ssl.set_host ssl host;
-            Lwt_ssl.ssl_perform_handshake s
+            with_socket (fun ssl ->
+                Ssl.set_client_SNI_hostname ssl host;
+                maybe_verify ssl (verification verify (hostname, ip)))
         | None -> Lwt_ssl.ssl_connect fd ctx)
```
hannesm, line 108: shouldn't here in the none case, when no hostname is provided, still the ip verificaton been done?

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
```diff
@@ -52,22 +52,59 @@ module Client = struct
...
   let default_ctx = create_ctx ()
 
-  let connect ?(ctx = default_ctx) ?src ?hostname sa =
+  type verify = { hostname : bool; ip : bool }
+
+  let default_verify = { hostname = true; ip = false }
+
+  let verification { hostname; ip } = function
+    | None, _ when hostname -> invalid_arg "impossible to verify hostname"
+    | _, None when ip -> invalid_arg "impossible to verify ip"
```
hannesm, line 61: are exceptions the way to go? shouldn't this code be executed as soon as possible (before opening a socket etc.)?

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
```diff
@@ -52,22 +52,59 @@ module Client = struct
...
+  let validate_hostname hostname =
+    try
+      let _ = Domain_name.(host_exn (of_string_exn hostname)) in
+      hostname
+    with Invalid_argument msg ->
+      let s =
+        Printf.sprintf "couldn't convert %s to a [`host] Domain_name.t: %s"
+          hostname msg
+      in
+      invalid_arg s
```
hannesm, line 76: This validate_hostname function does only ensure that the provided string is a `` [`host] Domain_name.t ``, this is its purpose?

src/conduit-lwt-unix/conduit_lwt_unix_ssl.dummy.mli
psafont: > I wonder, shouldn't ?verify be implicit? 

Unfortunately, no. hostname can easily be a string representing an IP, which makes the peers' certificates unverifiable.

Regarding IP, it's rare that certificates contain IP addresses while this call will also contain an IP when using cohttp if I understood it correctly, so it's going to be rare the case where it needs to be verified.

Since both verifications are needed case-by-casse bases I think it's better if it's explicit with a reasonable default. Having them implicit wouldn't work with xenserver's use-case

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
psafont: "hostname" is a misnomer, it is a string that represents a remote host, it can easily be an IP.

I made the option to keep the previous behaviour of sending it as an SNI.

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
psafont: Yes, hostname can contain an IP

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
psafont: I'll try to push the check to be done as soon as possible.

Exceptions are used elsewhere so I thought it was reasonable to keep using them

psafont: I've pushed this check before to the very beginning of connect, before any socket is created

psafont:
I've tried to clarify the confusing situation of "hostname"s by changing some the names of some values and explaining it into a commit message. Not sure what more can I do there without breaking compatibility

src/conduit-lwt-unix/conduit_lwt_unix_ssl.real.ml
hannesm: My remark was not about SNI or not (which, if `hostname` is an IP, shouldn't be set), but that in the `hostname = None` case there's no call to `maybe_verify` which I'd have expected, but only a call to `Lwt_ssl.ssl_connect` -- so what kind of verification (if any) is performed here?

Of course, if you're not interested in that code path, leave it as is. It would be great if you could formulate an entry for CHANGES: which behaviour changed and why.

psafont: > if hostname is an IP, shouldn't be set)

Will change it, I was being conservative and retaining previous behaviour

> there's no call to maybe_verify which I'd have expected

I see what you mean, I'll change it to check the IP if the setting is enabled, in general I don't think it should be done as few certificates contain an IP as the subject.

> It would be great if you could formulate an entry for CHANGES: which behaviour changed and why.

Sure thing

hannesm: > > if hostname is an IP, shouldn't be set)
>
> Will change it, I was being conservative and retaining previous behaviour

I don't know for sure how OpenSSL behaves, but in the TLS RFC it is very clear that setting the server name identification to an IP address is invalid. It must be a valid hostname. If the current patch works fine for you in such a scenario, feel free to leave it as is.

> I see what you mean, I'll change it to check the IP if the setting is enabled, in general I don't think it should be done as few certificates contain an IP as the subject.

Well, most DNS-over-TLS certificates include an IP address. But again, if you think it is not worth it, I'm fine to merge as is.

psafont: Now it shouldn't send the SNI when a DNS hostname isn't available and it should be able to verify the IP in the certificate if the users wants it to, let me know if the changelog looks good

psafont:
unit tests look good. I'll do a whole build and let you know tomorrow whether the tests went well

psafont:
All tests passed successfully :)

hannesm:
thanks for your patience, merged and released as 6.0.2 in https://github.com/ocaml/opam-repository/pull/22665

hannesm:
Sorry, but in the case that `ssl` is not installed (it is an optional dependency), the opam-repository-ci caught an error:

```
#=== ERROR while compiling conduit-lwt-unix.6.0.2 =============================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.08.1 | pinned(https://github.com/mirage/ocaml-conduit/releases/download/v6.0.2/conduit-6.0.2.tbz)
# path                 ~/.opam/4.08/.opam-switch/build/conduit-lwt-unix.6.0.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p conduit-lwt-unix -j 255
# exit-code            1
# env-file             ~/.opam/log/conduit-lwt-unix-7-dd95c6.env
# output-file          ~/.opam/log/conduit-lwt-unix-7-dd95c6.out
### output ###
# (cd _build/default && /home/opam/.opam/4.08/bin/ocamlc.opt -w -40 -g -bin-annot -I src/conduit-lwt-unix/.conduit_lwt_unix.objs/byte -I /home/opam/.opam/4.08/lib/angstrom -I /home/opam/.opam/4.08/lib/astring -I /home/opam/.opam/4.08/lib/base/caml -I /home/opam/.opam/4.08/lib/bigstringaf -I /home/opam/.opam/4.08/lib/conduit -I /home/opam/.opam/4.08/lib/conduit-lwt -I /home/opam/.opam/4.08/lib/domain-name -I /home/opam/.opam/4.08/lib/ipaddr -I /home/opam/.opam/4.08/lib/ipaddr-sexp -I /home/opam/.opam/4.08/lib/ipaddr/unix -I /home/opam/.opam/4.08/lib/logs -I /home/opam/.opam/4.08/lib/lwt -I /home/opam/.opam/4.08/lib/lwt/unix -I /home/opam/.opam/4.08/lib/macaddr -I /home/opam/.opam/4.08/lib/ocaml/threads -I /home/opam/.opam/4.08/lib/ocplib-endian -I /home/opam/.opam/4.08/lib/ocplib-endian/bigstring -I /home/opam/.opam/4.08/lib/parsexp -I /home/opam/.opam/4.08/lib/ppx_sexp_conv/runtime-lib -I /home/opam/.opam/4.08/lib/sexplib -I /home/opam/.opam/4.08/lib/sexplib0 -I /home/opam/.opam/4.08/lib/stringext -I /home/opam/.opam/4.08/lib/uri -I /home/opam/.opam/4.08/lib/uri/services -no-alias-deps -o src/conduit-lwt-unix/.conduit_lwt_unix.objs/byte/conduit_lwt_unix.cmi -c -intf src/conduit-lwt-unix/conduit_lwt_unix.pp.mli)
# File "src/conduit-lwt-unix/conduit_lwt_unix.mli", line 164, characters 11-22:
# 164 |   ?ssl_ctx:Ssl.context ->
#                  ^^^^^^^^^^^
# Error: Unbound module Ssl
```

I suspect we'll have to define `Conduit_lwt_unix_ssl.Client.ctx` (in _dummy as unit, in _real as an alias for Ssl.context) - does this make sense?

psafont:
Ah, sorry about that. Yes the solution makes sense. not sure how to do the testing to ensure it works fine

hannesm:
@psafont great, would you have time to provide a PR?

psafont:
Yes, of course. I'm now doing some sanity checking before opening it

PR #418, opened 2022-12-14
conduit-lwt-unix: do not use directly the Ssl module

psafont:
Instead create an alias to the types needed in conduit_lwt_unix_ssl modules and use the alias.
This makes it possible to compile conduit-lwt-unix without openssl


hannesm:
thanks.

PR #427, opened 2024-06-24
Switch to sexplib0 for lighter dependencies

emillon:
`sexplib0` only depends on `ocaml` and `dune`. It contains the type definition and some conversion functions. `conduit` does not use anything else (like a parser, etc) so it can use it.
(`sexplib` already depends on `sexplib0` so it's using the same types under the hood)

The corresponding bits are:
- `sexplib` (opam package and library) -> `sexplib0`
- `Sexplib.Conv` (module) -> `Sexplib0.Sexp_conv`
- `Sexplib.Std` -> `Sexplib0.Sexp_conv` (we only need these bits)
- `Sexplib.Sexp.to_string_hum` -> `Sexplib0.Sexp.to_string_hum`

### DIFF
--- a/conduit-async.opam
+++ b/conduit-async.opam
@@ -14,7 +14,7 @@ depends: [
   "uri" {>= "4.0.0"}
   "ppx_here" {>= "v0.9.0"}
   "ppx_sexp_conv" {>="v0.13.0"}
-  "sexplib"
+  "sexplib0"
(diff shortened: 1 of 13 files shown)


MisterDA (approved):

hannesm:
Thanks for your contribution. The CI errors are due to other packages got updated in the meantime.

PR #428, opened 2024-07-18
TLS tunnel over Lwt_io.channel

art-w:
I'm marking this PR as a draft, because it exposes the functionalities of TLS-over-TLS added by this PR https://github.com/mirleft/ocaml-tls/pull/499 which hasn't yet been reviewed... but I have some questions on how to best present this feature in Conduit :)

For context, @MisterDA has a working implementation of `cohttp-lwt-unix` client requests with http/https proxying, which relies on `conduit-lwt-unix` for TLS and on this PR for the "https over http(s)", which happens because https proxies provide a safe tunnel for the client to negotiate the TLS connection with the server. For this to work though, we need to able to reuse an existing connection instead of having Conduit create a new socket. The https://github.com/mirleft/ocaml-tls/pull/499 PR has more details on how this works, but I've also added an example in Conduit to test the flow if you are curious.

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_unix.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_unix.ml
@@ -141,7 +141,7 @@ type vchan_flow = { domid : int; port : string } [@@deriving sexp]
 
 type flow =
   | TCP of tcp_flow
-  | Tunnel
+  | Tunnel of string * ic * oc
   | Domain_socket of domain_flow
   | Vchan of vchan_flow
 [@@deriving sexp]
@@ -294,10 +294,10 @@ let connect_with_tls_native ~ctx (`Hostname hostname, `IP ip, `Port port) =
 
 let connect_with_tls_tunnel ~ctx (`Hostname hostname, ic, oc) =
   certificates ~ctx >>= fun certificates ->
-  let hostname = domain_name hostname in
+  let host = domain_name hostname in
   Conduit_lwt_tls.Client.tunnel ?certificates
-    ~authenticator:ctx.tls_authenticator hostname (ic, oc)
-  >|= fun (ic, oc) -> (Tunnel, ic, oc)
+    ~authenticator:ctx.tls_authenticator host (ic, oc)
+  >|= fun (ic', oc') -> (Tunnel (hostname, ic, oc), ic', oc')
 
 let connect_with_openssl ~ctx (`Hostname host_addr, `IP ip, `Port port) =
   let sa = Unix.ADDR_INET (Ipaddr_unix.to_inet_addr ip, port) in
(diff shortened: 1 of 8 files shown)


src/conduit-lwt-unix/conduit_lwt_unix.ml
```diff
@@ -410,6 +423,7 @@ let serve ?backlog ?timeout ?stop ~on_exn ~(ctx : ctx) ~(mode : server) callback
 
 let endp_of_flow = function
   | TCP { ip; port; _ } -> `TCP (ip, port)
+  | Tunnel -> `Unknown "TLS tunnel"
```
art-w, line 426: I'm not sure what is the best way to address this? We can't expose `Lwt_io.channel` in `Conduit.endp` obviously, should `conduit-lwt-unix` expose an extended `endp` type with the new tunnel option?

hannesm:
What state is this in? Is there anyone interested in conduit reviewing feature PRs? I still don't understand conduits purpose, and only do maintenance (mainly because I want to release mirage) ;)

art-w:
I've rebased and removed the draft status, since the tls dependency has since been released :)
The feature has been thoroughly tested with https://github.com/mirage/ocaml-cohttp/pull/1080 so I think it's in good shape, although it would be nice to have a conduit expert opinion on the `endp_of_flow` question.

src/conduit-lwt-unix/conduit_lwt_unix.ml
art-w: I went ahead and exposed a new `Conduit_lwt_unix.endp` which extends `Conduit.endp` with the Lwt-specific TLS tunneling

art-w:
I went with exposing the TLS tunneling over an existing `Lwt_io.channel` connection by exposing a new type `Conduit_lwt_unix.endp` which extends `Conduit.endp` with the Lwt-specific feature, as this seemed like the most natural solution... @samoht Would you have some time to look at this PR? :)

src/conduit-lwt-unix/conduit_lwt_unix.ml
```diff
@@ -262,31 +264,41 @@ let set_max_active maxactive = Conduit_lwt_server.set_max_active maxactive
...
+      >|= fun certificate -> Some (`Single certificate)
+
+let domain_name hostname =
+  try Domain_name.(host_exn (of_string_exn hostname))
+  with Invalid_argument msg ->
+    let s =
+      Printf.sprintf "couldn't convert %s to a [`host] Domain_name.t: %s"
+        hostname msg
+    in
+    invalid_arg s
```
MisterDA, line 283: You can preserve the backtrace of the exception if you intent to re-raise (even with a modified message) it:

```suggestion
    let trace = Printexc.get_raw_backtrace () in
    let msg =
      Printf.sprintf "couldn't convert %s to a [`host] Domain_name.t: %s"
        hostname msg
    in
    Printexc.raise_with_backtrace (Invalid_argument msg) trace
```

MisterDA (approved):

samoht:
LGTM

PR #429, opened 2024-07-21
Update GitHub Actions

smorimoto:
https://github.com/smorimoto/ocaml-conduit/actions/runs/10027482004


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
samoht, line 1: why is it useful to run dependabot here?

samoht:
Thanks!

PR #430, opened 2024-07-22
Raise and reraise exceptions with `Stdlib` rather than `Lwt`.

MisterDA:
Lwt's documentation reads:

> In most cases, it is better to use `failwith s` from the standard
> library.

and

> Whenever possible, it is recommended to use `raise exn` instead, as
> raise captures a backtrace, while `Lwt.fail` does not. If you call
> `raise exn` in a callback that is expected by Lwt to return a
> promise, Lwt will automatically wrap `exn` in a rejected promise,
> but the backtrace will have been recorded by the OCaml runtime.
>
> For example, `bind`'s second argument is a callback which returns a
> promise. And so it is recommended to use `raise` in the body of that
> callback.
>
> Use `Lwt.fail` only when you specifically want to create a rejected
> promise, to pass to another function, or store in a data structure.

Prefer to capture backtraces to improve debugability.

See also [More raise less fail #1008](https://github.com/ocsigen/lwt/pull/1008).

### DIFF
--- a/src/conduit-lwt-unix/conduit_lwt_launchd.dummy.ml
+++ b/src/conduit-lwt-unix/conduit_lwt_launchd.dummy.ml
@@ -15,4 +15,4 @@
  *
  *)
 
-let activate _fn _name = Lwt.fail_with "No Launchd support"
+let activate _fn _name = failwith "No Launchd support"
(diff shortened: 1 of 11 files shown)


hannesm:
Thanks for your contribution. I merged the main branch into your PR, and will merge this now.

Issue #420, opened 2023-02-28
Optional dependencies concidered harmful


kit-ty-kate:
The change in https://github.com/mirage/ocaml-conduit/pull/419 is causing nasty problems at runtime everywhere (users upgrading to tls >= 0.16.0 and trying to run software that previously expected a dependency toward `tls` to be enough for conduit to use/enable tls at all)

In my opinon, implicit/optional dependencies are harmful to any ecosystem all users and should only be used when strictly necessary, as a last resort. In the case of Conduit, I'm not versed in how it is designed but it feels like the following can be done:
- Split `conduit-lwt-unix`'s optional dependencies into 3 new sub-packages (`conduit-lwt-unix-tls`, `conduit-lwt-unix-ssl`, `conduit-lwt-unix-launchd`)
- Add a common semi abstract type in `Conduit_lwt_unix` so this module knows how deal with the TLS/SSL backend chosen
- Add a function returning this new type containing all the necessary endpoint that the backend agnostic `Conduit_lwt_unix` module requires.

avsm:
Before we consider solutions, what's actually gone wrong for our users after #419?  Is it that they now need to do `opam install cohttp-lwt-unix tls-lwt`, whereas previously it was just `opam install cohttp-lwt-unix tls` ?

The original historical reason for supporting both OpenSSL and OCaml TLS in conduit is that the latter wasn't mature enough. But now, the reverse is true: the ssl bindings seem to be less well maintained than OCaml TLS.  I'm tempted to suggest dropping support for the OpenSSL in Conduit -- perhaps we should canvas opinions on this from discuss.ocaml.org.

kit-ty-kate:
> Before we consider solutions, what's actually gone wrong for our users after #419? Is it that they now need to do `opam install cohttp-lwt-unix tls-lwt`, whereas previously it was just `opam install cohttp-lwt-unix tls` ?

yes, or more usually this is already in the chain of dependencies so people don't have to think about it. Similarly `tls.lwt` was almost always present because most people have already `lwt` in their dependencies but since it has been moved to its own library `tls-lwt` (that's a good thing) this breaks (sometime unknowingly) users of this sublibrary. See https://github.com/ocaml/opam-repository/pull/23344#issuecomment-1439945050 for two examples in sihl and current_gitlab, ore more recently in ocurrent https://github.com/ocurrent/ocurrent/pull/414 where CI systems that use it just crashed at runtime because no TLS backend was present anymore.

[...]

PR #431, opened 2024-08-29
Improve error message for missing SSL/TLS support 

### LINKED ISSUE #423: Improve error message `No SSL or TLS support compiled into Conduit`
I've received this error when using cohttp and used a variety of bad solution to work around this error (as described [here](https://discuss.ocaml.org/t/no-ssl-or-tls-support-compiled-into-conduit-with-cohttp-5-3-1-conduit-6-2-1-tls-0-17-4/14423/1)). 
However, the cohttp README.md actually specifes the correct solution: installing tls-lwt or lwt_ssl. 

Would it be possible to improve the error message in conduit to make this more obvious? Or perhaps this is rather an issue in cohttp? Alternatively, some additional instructions in the README of Conduit might also help users.

filipeom:
Closes #423 

I don't know if this is makes sense due to #420. But it could potentially save a google search :sweat_smile:


hannesm:
Thanks for your contribution!

PR #432, opened 2024-08-29
Adapt to tls 1.0.0 API changes

hannesm:


art-w (approved):
Looks good! (I ran into the same breaking change and was about to submit the same fixes so the review was easy ^^)

hannesm:
I rebased on top of the main branch and will wait for CI here.

